PR #3, opened 2016-10-31
USAspending Frontend: Linting

kevinli-work:
* Added Airbnb linting
* Fixed a lot of files to pass linting


.eslintrc
```diff
@@ -0,0 +1,46 @@
...
+    "extends": "airbnb",
+    "env": {
+        "es6": true
+    },
+    "globals": {
+        "window": true,
+        "document": true
+    },
+    // overrides to the Airbnb style follow
+    "rules": {
```
mab6bf: Very good work on these overrides - agree with all

PR #4, opened 2016-11-02
DS-396 Feature search ui

ejgullo:
- Created UI portion of Award Type filter
- Included default values until API is connected
- Primary and Secondary tier Award Types supported


src/js/components/search/filters/awardType/PrimaryAwardType.jsx
```diff
@@ -0,0 +1,52 @@
...
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            showSubItems: false
+        };
+    }
+
+    toggleSubItems() {
+        if (this.state.showSubItems === false) {
```
kevinli-work: Could we replace this logic with something even simpler like:

```
this.setState({
    showSubItems: !this.state.showSubItems
});
```

src/js/components/search/filters/awardType/PrimaryAwardType.jsx
```diff
@@ -0,0 +1,52 @@
...
+            this.setState({ showSubItems: true });
+        }
+        else {
+            this.setState({ showSubItems: false });
+        }
+    }
+
+    render() {
+        const primaryAwardList = (<div className="primaryAwardTypeOption">
```
kevinli-work: Let's refactor this so we don't have a large chunk of HTML outside of the `render()` function. If you need to put this much HTML in, you may want to consider creating this as its own component.

```
let primaryAward = <CollapsedAwardType />
if (this.state.showSubItems) {
    primaryAward = ......something to show list instead
}
```

src/js/components/search/filters/awardType/SecondaryAwardType.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+const propTypes = {
+    subList: React.PropTypes.array
+};
+
+export default class SecondaryAwardType extends React.Component {
+
+
+    render() {
+        const secondaryAwardList = (this.props.subList.map((subList, index) =>
```
kevinli-work: Too many parens is actually counterproductive re: readability.
I would move the array iteration logic up to PrimaryAwardType to match how the list of PrimaryAwards is iterated through in its parent AwardType component. This component should just be responsible for rendering a _specific_ secondary award type, not the full list.

src/_scss/custom.scss
```diff
@@ -6,7 +6,9 @@
 
 // ***** PAGES
 @import './pages/homepage';
-@import './pages/searchpage';
+@import './pages/search/searchPage';
+@import './pages/search/searchSidebar';
```
kevinli-work: Let's move the import statements for search page components inside `searchPage.scss` so that those styles are constrained to just the search page, rather than being universal across the site. Universal styles were what caused the Sass to become unmaintainable last time.

src/js/components/search/filters/awardType/PrimaryAwardType.jsx
```diff
@@ -23,29 +22,31 @@ export default class PrimaryAwardType extends React.Component {
...
-                <Icons.AngleDown />
-            </a>
-            {this.state.showSubItems === true ?
-                <SecondaryAwardType subList={this.props.subList} /> : null }
-        </div>);
+        let primaryAward = (<CollapsedAwardType
+            name={this.props.name}
+            click={this.toggleSubItems.bind(this)} />);
+
+        const secondaryAwardTypes = this.props.subList.map((subList, index) =>
```
kevinli-work: Your `render` function needs to be as light as possible because of how frequently it may be executed. In the current logic, you are always looping through the secondary award types, regardless of whether or not you need to (ie, the filter is collapsed). But if you move the loop inside of the `if` block, it only runs on demand.

src/js/components/search/filters/awardType/SecondaryAwardType.jsx
```diff
@@ -6,22 +6,19 @@
...
-            </div>));
-
         return (
-            <div>
-                { secondaryAwardList }
+            <div key={this.props.key} className="secondaryAwardTypeOption subList">
+                <input
+                    type="checkbox"
+                    id={this.props.subListValue} value={this.props.subListValue} />
+                <label htmlFor={this.props.subListValue}>{this.props.subListValue}</label>
```
kevinli-work: `this.props.subListValue` may return normal strings like "Capitalized Word With Spaces", which are not valid HTML `id` or `htmlFor` formats. Let's think of a way that will pass an appropriate value for `id` and `htmlFor`.

PR #5, opened 2016-11-07
Feature search ui

ejgullo:
-DS-396, 397
- Bug fixes as per email
- Restructuring render in Primary Awards to lighten render
- Adding Fiscal Year UI components
- Linter fixes across search UI components


kevinli-work:
A few things need to happen before we can start this PR:

- [ ] Finish and merge Redux branch
- [ ] Use Redux branch's prop data structure

src/_scss/pages/search/awardTypeFilter.scss
```diff
@@ -29,6 +26,10 @@
         top: .25rem;
     }
 
+    .primaryAwardTypeOption label div, .secondaryAwardTypeOption label div {
```
kevinli-work: FYI (for the future), you should take advantage of Sass's nesting for more readable and easier to develop code:
```
.primaryAwardTypeOption, .secondaryAwardTypeOption {
    label {
        &:before {
            top: .25rem;
        }
        div {
            display: inline;
        }
    }
}
```

This makes it easier to read and think about.

src/js/components/search/filters/timePeriod/TimePeriod.jsx
```diff
@@ -0,0 +1,145 @@
...
+    label: React.PropTypes.string
+};
+
+export default class TimePeriod extends React.Component {
+
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            state: 'loading',
```
kevinli-work: What are we using `state.state` for?

src/js/components/search/filters/timePeriod/FiscalYear.jsx
```diff
@@ -0,0 +1,23 @@
...
+const propTypes = {
+    name: React.PropTypes.string
+};
+
+export default class FiscalYear extends React.Component {
+
+    render() {
+        return (
+            <div className="fiscalYearOption">
+                <input type="checkbox" id={this.props.name} value={this.props.name} />
```
kevinli-work: Again, let's make the HTML IDs valid (no spaces). Maybe just have the parent provide the year and prepend the "FY" string in this component's `label` and them building a compliant HTML ID with that.

Review ES6's [template strings feature](https://developers.google.com/web/updates/2015/01/ES6-Template-Strings#string_substitution).

src/js/components/search/filters/timePeriod/DateRange.jsx
```diff
@@ -0,0 +1,52 @@
+/**
+ * DateRange.jsx
+ * Created by Emily Gullo 11/03/2016
+ **/
+
+import React from 'react';
+import DatePicker from '../../../sharedComponents/DatePicker';
```
kevinli-work: With the latest `dev` update, we can now provide import paths that use `src/js` as the base directory. So this would be `import Datepicker from 'components/sharedComponents/DatePicker';`

You can still reference relative to the current file when that is more efficient. Start *relative* imports with `./`, `../`, etc. Don't use leading slashes or dots when importing relative to the base source directory.

src/_scss/pages/search/timePeriodFilter.scss
```diff
@@ -0,0 +1,38 @@
...
+        max-height:1.5rem;
+    }
+
+    .fiscalYearOption label::before {
+        height:1.2rem;
+        width:1.2rem;
+    }
+
+    .fiscalYearOption:nth-child(odd){
+        float:left;
```
kevinli-work: We have Bourbon and Neat. Can we use the grid system instead of floating ourselves? Also why are we floating inside a `display: table-cell`?

src/js/components/search/filters/timePeriod/TimePeriod.jsx
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    validateDates() {
+        // validate that dates are provided for both fields and the end dates
+        // don't come before the start dates
+
+        const rng = Object.assign({}, this.state);
+
+        const output = {
+            range: rng
```
kevinli-work: What is `range`?

src/js/components/search/filters/timePeriod/TimePeriod.jsx
```diff
@@ -0,0 +1,145 @@
...
+                header: '',
+                description: ''
+            })
+        });
+    }
+
+
+    render() {
+        const fiscalYears = this.props.timePeriods.map((year, index) =>
+            <FiscalYear name={year} key={index} />
```
kevinli-work: `FiscalYear` might make more sense semantically as a `ul` list under the hood than a series of `div`s (you can, of course, put `div`s inside each list item).

src/js/components/search/filters/timePeriod/TimePeriod.jsx
```diff
@@ -57,10 +56,10 @@ export default class TimePeriod extends React.Component {
         // validate that dates are provided for both fields and the end dates
         // don't come before the start dates
 
-        const rng = Object.assign({}, this.state);
+        const err = Object.assign({}, this.state);
 
         const output = {
-            range: rng
+            error: err
```
kevinli-work: I still don't understand what this is happening here. In the `constructor`, the state is defined as having a child object named `error`. But here, you're cloning a new instance of the state, and then setting the value of `output`'s `error` key to the _entire_ cloned state...which incidentally, gets overwritten anyway in lines 68-95.

src/js/components/search/filters/timePeriod/TimePeriod.jsx
```diff
@@ -0,0 +1,144 @@
...
+                show: false,
+                header: '',
+                description: ''
+            };
+        }
+
+        this.setState(output);
+    }
+
+    showError(head, desc) {
```
kevinli-work: So.............what does `showError` and `hideError` actually do?

src/js/components/search/filters/timePeriod/TimePeriod.jsx
```diff
@@ -34,15 +22,20 @@ export default class TimePeriod extends React.Component {
         this.state = {
```
kevinli-work: What about fiscal years? The top filter component's state doesn't seem to store that information/it's not bubbled up from the child components.

src/js/components/search/filters/timePeriod/DateRange.jsx
```diff
@@ -14,35 +15,58 @@ const defaultProps = {
...
-                        <div className="through-text">
-                            through
-                        </div>
-                        <DatePicker
-                            type="endDate" title="End Date" tabIndex={this.props.startingTab + 4}
-                            onDateChange={this.props.onDateChange} value={this.props.endDate}
-                            opposite={this.props.startDate} showError={this.props.showError}
-                            hideError={this.props.hideError} />
-                    </div>
+                    <DatePicker
```
kevinli-work: Can we keep this from happening (datepicker being off screen)? 
screen shot 2016-11-14 at 11 05 07 am

src/_scss/pages/search/timePeriodFilter.scss
```diff
@@ -55,6 +57,42 @@
             @include usa-input;
             display:inline;
         }
+        .date-range-wrapper {
+
+            .errorMessage {
```
kevinli-work: This style is not applied because `errorMessage` is not a child class of `date-range-wrapper`. Also let's start using `dashed-names` instead of camelCase.

src/js/components/search/filters/timePeriod/TimePeriod.jsx
```diff
@@ -94,7 +129,12 @@ export default class TimePeriod extends React.Component {
         }
 
         if (this.state.shownFilter === 'fy') {
```
kevinli-work: Why is this in here twice? Almost identical code is in lines 117-124. What is the case that this `if` is handling that isn't handled there? If this logical structure is necessary, is there a way to make this less repetitive?

src/js/components/search/filters/timePeriod/AllFiscalYears.jsx
```diff
@@ -58,15 +40,12 @@ export default class AllFiscalYears extends React.Component {
...
         }
 
-        this.setState({
-            selectedFY: arrayFY,
-            allFY: allSelected
-        });
+        this.props.saveSelected(arrayFY, allSelected);
     }
 
     saveAllYears() {
```
kevinli-work: There is a bug in this function. If some FY years are individual selected before the user clicks "All Years," the All Years checkbox is checked but all the individual year entries become unchecked. Subsequently, the All Years checkbox is always inverse from what is checked (checked if the other years are unchecked, unchecked if the other years are checked).

I don't think this function needs to consider what values are already checked. We know if the user is going from unchecked -> checked or vice versa via the `allFY` props. Given the user's action, we should just completely overwrite the `selectedFY` values with either the full list of FYs or an empty set. We already have both on-hand, so no merges should be necessary (esp since the merge operation still outputs a new instance per `immutability-js`).

src/js/components/search/filters/timePeriod/TimePeriod.jsx
```diff
@@ -82,9 +98,28 @@ export default class TimePeriod extends React.Component {
...
     }
 
+    saveSelected(arrayFY, allSelected) {
+        this.setState({
+            selectedFY: arrayFY,
+            allFY: allSelected
+        });
+    }
+
+    saveAll(arrayFY, allFY) {
```
kevinli-work: Do we need this function? Could we get rid of it and use `saveSelected` instead?

ejgullo:
So strange, locally they are parent child, even on a restart of the process.  Moved it up anyhow - they really should be siblings.  Maybe I'm missing something from dev?  Not sure.  Also added a check on setting errorDetails up to ensure shownFilter is set to Date Range.

PR #8, opened 2016-11-16
Page Layout

kevinli-work:
* Restructuring Sass and site body components for semantic HTML (supporting 508 compatibility) and style layout per designs
* Updates to award type filter to align with designs and correctly word wrap


src/_scss/layouts/default/header/_positioning.scss
```diff
@@ -0,0 +1,27 @@
+@mixin headerPositioning {
+    .nav-container {
+        @include outer-container();
+        max-width: 100%;
+
+        @include pad(1rem 0);
+        @include media($nav-width) {
+            // @include pad(1rem 6rem);
```
mab6bf: Do we need this? If not, let's remove

PR #9, opened 2016-11-18
Feature search ui

ejgullo:
- Adding functionality to toggle filters open and closed


kevinli-work:
Can you fix the collapsed state to look like the designs without all the extra padding?
screen shot 2016-11-21 at 9 30 23 am

kevinli-work:
The calendar icons need to be aligned:
screen shot 2016-11-21 at 9 32 22 am

The calendars themselves should be below, not overlapping, the input fields:
screen shot 2016-11-21 at 9 34 06 am

src/js/components/search/filters/awardType/AwardType.jsx
```diff
@@ -39,19 +40,32 @@ const defaultProps = {
...
+                    <PrimaryAwardType {...type} {...this.props} key={index} />
+            ));
+        }
+
 
         return (
-            <div className="award-type-filter">
-                <b>Award Type</b>
+            <div className="award-type-filter search-filter">
+                <FilterExpandButton
```
kevinli-work: Why are the `FilterExpandButton` components in every filter component? Shouldn't these, plus the title of the filter, be in the parent component?

src/js/components/search/filters/timePeriod/TimePeriod.jsx
```diff
@@ -143,23 +148,34 @@ export default class TimePeriod extends React.Component {
             activeClassDR = '';
         }
 
-        return (
-            <div className="time-period-filter">
-                <b>Time Period</b>
+        let timePeriod = null;
+        if (this.props.showFilter === true) {
+            timePeriod = (<div>
```
kevinli-work: This is too much JSX in the logic; refactor this code so the bulk of the JSX is in the return rather than in a variable. This may involve rethinking how you've restructured your components.

src/_scss/pages/search/searchSidebar.scss
```diff
@@ -3,16 +3,62 @@
...
 		margin-bottom:1rem;
 
         label {
             @include usa-label;
         }
-	}
 
-	.search-option b::after {
-		content: ' \02193';
+        svg {
```
kevinli-work: Since this is for a specific component (the filter expand/collapse button), this should be in its own file.

src/_scss/pages/search/searchSidebar.scss
```diff
@@ -3,16 +3,62 @@
...
+            fill: $color-gray;
+            height: 2rem;
+            width: 2rem;
+            position:relative;
+            cursor: pointer;
+        }
+
+        .toggle {
```
kevinli-work: Don't just copy-paste the Sass, rename and/or otherwise constrain the style so it won't conflict or be affected by changes to the source it was copied from  (the award subtype button).

kevinli-work:
The filters should probably be expandable/collapsable by clicking on the title text as well, not just the button.
screen shot 2016-11-21 at 11 10 21 am

kevinli-work:
Can you fix the margins on the date error message so it is equal on both sides?
screen shot 2016-11-21 at 11 10 57 am

src/js/components/search/filters/awardType/AwardType.jsx
```diff
@@ -41,35 +40,26 @@ const defaultProps = {
...
-                this.props.awardTypes.map((type, index) =>
-                    <PrimaryAwardType {...type} {...this.props} key={index} />
-            ));
-        }
+        const awardTypes = (
+            this.props.awardTypes.map((type, index) =>
+                <PrimaryAwardType {...type} {...this.props} key={index} />
+        ));
 
+        if (this.props.showFilter === false) {
```
kevinli-work: Why not put the show/hide logic in the parent component?

src/js/components/search/SearchOption.jsx
```diff
@@ -39,27 +40,32 @@ export default class SearchOption extends React.Component {
...
+            case 'Time Period':
+                searchOption = (<TimePeriod
+                    showFilter={this.state.showFilter} />);
+                break;
             default:
-                return (<div className="search-option"><b>{this.props.name}</b></div>);
+                searchOption = null;
         }
+
+        return (<div className="search-option">
```
kevinli-work: Add some line breaks between the parens to match our standard coding style for component `render` return statements.

src/_scss/pages/search/searchSidebar.scss
```diff
@@ -3,64 +3,19 @@
...
-            border-bottom: 1px solid $color-gray-lighter;
-
-            .filter-header {
-                padding-left: 0;
-            }
-        }
-	}
-
     @import "./_filters";
+    @import "./searchOption";
```
kevinli-work: What is the convention you're using for determining whether a Sass file should be `_`-prefixed or not?

kevinli-work: My personal approach is to use `_` prefixes for files that support a section (a file that just deals with positioning as `_positioning.scss` for example) while no `_` for files that are the "entrypoint" when being imported by external files (`sidebar` for example, which in turn imports `_positioning.scss`).

src/_scss/components/header/_headerSubMenu.scss
```diff
@@ -0,0 +1,43 @@
...
+  @include media($nav-width) {
+    @include unstyled-list;
+    @include padding(1.15rem null);
+    background-color: $color-primary-darkest;
+    min-width: 21.5rem;
+    width: auto;
+    position: absolute;
+  }
+
+  &[aria-hidden=true] {
```
kevinli-work: Is this 18F code or ours? We should not be Aria tags should be descriptive only, not be used to describe.

kevinli-work: I see Aria being used elsewhere in a similar way; if this our code, we should rethink its use.

src/_scss/components/footer/_footerSocial.scss
```diff
@@ -0,0 +1,54 @@
+@mixin usa-social_link {
```
kevinli-work: Why social_link and not social-link?

kevinli-work: Also for every other instance of similarly named classes/mixins

src/_scss/pages/search/searchSidebar.scss
ejgullo: I was vaguely attempting to go with what you did - I think I misunderstood it.  Renaming a few files, but let me know if I'm still not quite on.

PR #10, opened 2016-11-21
Redux search results store

kevinli-work:
* Store search results as a flat data structure in Redux for performance reasons
* Created model objects to represent each record


kevinli-work:
Must be merged **before** #11

PR #22, opened 2016-12-16
Sidebar UI

ejgullo:
Once more with feeling!
• Updating sidebar to adjust at 4 breakpoints (this will be updated in future stories)
• Updating Time Period filters' items to fit as the sidebar resizes


src/_scss/lib/neat/settings/_grid.scss
```diff
@@ -3,7 +3,7 @@
 $column: em(46.875) !default;
 $gutter: em(30) !default;
 $grid-columns: 16 !default;
-$max-width: em(1200) !default;
+$max-width: em(1400) !default;
```
kevinli-work: Let's not make changes to the `lib` files. You should be able to override these settings in the `core` directory.

PR #25, opened 2016-12-19
Location Filter

mab6bf:


kevinli-work:
Should the autocomplete attempt to complete again when the user toggles the all/domestic/foreign option? Doing so defocuses the input, which causes the awesomplete to go away (and also no HTTP call is made).

src/js/redux/reducers/search/filters/locationFilterFunctions.js
```diff
@@ -0,0 +1,20 @@
...
+        updatedSet = updatedSet.delete(value);
+    }
+    else if (!updatedSet.includes(value) && direction === 'add') {
+        updatedSet = updatedSet.add(value);
+    }
+
+    return updatedSet;
+};
+
+export default updateSelectedLocations;
```
kevinli-work: Can we change this to not be a `default` export? I understand the linter will give you a warning because there is only one export in the file, but this will:

1. allow us to add more exported functions in the future without changing the `import` syntax in other files
2. be consistent with how other plain JS files that export functions behave

If you can't handle the linter warnings, you can suppress them using:
```
/* eslint-disable import/prefer-default-export */
// provide an explanation - required!
your code
/* eslint-enable import/prefer-default-export */
```
Make sure you re-enable after your code block!

src/js/redux/reducers/search/searchFiltersReducer.js
```diff
@@ -6,13 +6,15 @@
 import { Set } from 'immutable';
 
 import * as AwardFilterFunctions from './filters/awardFilterFunctions';
+import updateSelectedLocations from './filters/locationFilterFunctions';
```
kevinli-work: See previous comment. Let's change this to a namespaced wildcard `import` to be consistent with how other such imports are handled in the codebase.

src/js/redux/actions/search/searchFilterActions.js
```diff
@@ -31,3 +31,14 @@ export const updateGenericFilter = (state) => ({
...
     type: 'CLEAR_SEARCH_FILTER_ALL'
 });
+
+export const updateSelectedLocations = (state) => ({
+    type: 'UPDATE_SELECTED_LOCATIONS',
+    location: state.location,
+    direction: state.direction
+});
+
+export const setAutocompleteLocations = (state) => ({
```
kevinli-work: Is it consistent to put the autocomplete action in this file when it is generally handled in separate autocomplete files elsewhere in the Redux implementation?

mab6bf:
@kevinli-work Absolutely - I'll work on adding that tomorrow

src/js/redux/reducers/search/filters/locationFilterFunctions.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Created by michaelbray on 12/12/16.
+ */
+
+import { Set } from 'immutable';
+
+const updateSelectedLocations = (state, value, direction) => {
+    let updatedSet = new Set(state);
+
+    if (updatedSet.includes(value) && direction === 'remove') {
```
kevinli-work: Is `direction` necessary in either of these `if` blocks (and also in the action, etc.)? It seems to me that we can just check for membership in the set and if it exists, remove, otherwise add.

src/js/redux/reducers/search/autocompleteReducer.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Created by michaelbray on 12/12/16.
+ */
+
+const initialState = [];
+
+const autocompleteReducer = (state = initialState, action) => {
+    switch (action.type) {
+        case 'SET_AUTOCOMPLETE_LOCATIONS': {
+            return Object.assign([], action.locations);
```
kevinli-work: I'm not 100% sure you can do an `Object.assign` to clone an array. Lodash has a utility function for cloning an array. Alternatively, use an ImmutableJS structure (which we may migrate toward completely for Reducers in the future).

src/js/components/search/filters/location/CountryType.jsx
```diff
@@ -0,0 +1,44 @@
...
+    render() {
+        return (
+            <div className="location-radio">
+                <p>Show Only:</p>
+                <input
+                    type="radio"
+                    id="location-all"
+                    name="location"
+                    value="all"
+                    defaultChecked="true"
```
kevinli-work: Can the radio buttons be controlled by the LocationSearchContainer? Controlled components have their selection/checked value set by a parent component's state. When a change event occurs, it is bubbled up to the controlling component (as you currently do), but that parent component then pushes down the new selection/check value to the radio button, rather than allowing the browser to set it natively.

Take a look at the refactored award type filter components for an example of how it's done. Controlled components are important because there are multiple places where the filters can be set (such as the sticky top filter bar). By controlling selection and input values centrally, they can all remain in sync.

src/js/components/search/filters/location/LocationSearch.jsx
```diff
@@ -0,0 +1,40 @@
+/**
+ * LocationSearch.jsx
+ * Created by Emily Gullo 11/28/2016
+ **/
+
+import React from 'react';
+import CountryType from './CountryType';
+import SelectedLocations from './SelectedLocations';
+import LocationListContainer from '../../../../containers/sharedContainers/LocationListContainer';
```
kevinli-work: Let's change this import to use the new path resolving set up a few sprints ago. You can specify paths relative to the `/src/js` directory if you ditch the leading `./` or `../`.

src/js/components/search/filters/location/PlaceOfPerformanceTypeahead.jsx
```diff
@@ -0,0 +1,175 @@
+/**
```
kevinli-work: Let's keep a consistent naming scheme of `components/..../xxx.jsx` being the immediate child of `containers/.../xxxContainer.jsx`

src/js/containers/sharedContainers/LocationListContainer.jsx
```diff
@@ -0,0 +1,102 @@
...
+            itemLabel += ` in ${item.parent}`;
+        }
+
+        return {
+            label: itemLabel,
+            value: item.place
+        };
+    }
+
+    handleTextInput(locationInput) {
```
kevinli-work: Can we add 2 features to this:

1. There should be a timeout before a new API call goes out. Basically, if the user types really fast, we don't need to send out an API request on each keystroke since these are "expensive"; we can wait a few ms for the user to calm down first. Some timeout logic does exist in the child PlaceOfPerformanceTypeahead container, but that only deals with showing and hiding length error messages.

2. Even with a timeout, we need to ensure that we cancel any in-flight requests before starting a new one. API calls aren't guaranteed to finish in the order they were made, so an earlier request could return _after_ a later request, resulting in nonsensical data and wiping out the better data. For example, the user types `san `, an API call goes out (API call 1). The user keeps typing `san f`, second API call goes out (API call 2). If API call 2 comes back before API call 1, API call 1's callback/promise chain will cause the autocomplete array in Redux to be nuked and populated with API call 1's results (which may include results that no longer make sense, like `San Diego` and `San Antonio`).

src/js/containers/sharedContainers/LocationListContainer.jsx
```diff
@@ -0,0 +1,102 @@
...
+                    const data = res.data;
+                    let autocompleteData = [];
+
+                    // Filter out selectedLocations that may be included in the result
+                    if (this.props.selectedLocations.size > 0) {
+                        const locs = this.props.selectedLocations.toArray();
+
+                        data.forEach((loc) => {
+                            let isSelectedLocation = false;
+                            locs.forEach((selectedLoc) => {
```
kevinli-work: It seems not great do a complete loop through the selected locations array for every single item in the API response. Remember that Immutable (and also ES6's native equivalents) provide new data structures that allow you to test for membership without having to do the whole loop.

At minimum, you should use a `for` loop instead of `forEach` because you can exit that loop early (since we don't care about the remaining data once we make the first match).

kevinli-work: Alternatively, consider looking at Lodash's utility functions to see if there's a utility function that can do array comparisons/set operations for you.

src/_scss/pages/search/filters/location/_typeahead.scss
```diff
@@ -0,0 +1,30 @@
...
+
+    input {
+        font-size: $small-font-size;
+        height: 4rem;
+        margin: 0 auto;
+        padding:  11px 8px 11px 8px;
+        width: 90%;
+    }
+
+    .awesomplete ul li {
```
ejgullo: Working on this one~

src/js/components/search/filters/location/CountryType.jsx
ejgullo: Working on this one~

src/js/components/search/filters/location/LocationSearch.jsx
ejgullo: Working on this one~

src/js/components/search/filters/location/PlaceOfPerformanceTypeahead.jsx
ejgullo: Working on this one~

src/js/components/search/filters/location/SelectedLocations.jsx
```diff
@@ -0,0 +1,31 @@
...
+    selectedLocations: React.PropTypes.object,
+    removeLocation: React.PropTypes.func
+};
+
+export default class SelectedLocations extends React.Component {
+    render() {
+        const shownLocations = this.props.selectedLocations.map((location, key) => (
+            <ShownLocation
+                location={location}
+                id={key}
```
ejgullo: Working on this one~

src/js/redux/actions/search/searchFilterActions.js
mab6bf: I'll move to a separate file

src/js/redux/reducers/search/filters/locationFilterFunctions.js
mab6bf: Yep, I'll make that change

src/js/redux/reducers/search/searchFiltersReducer.js
mab6bf: Yep, I'll make this along with the previous

src/js/redux/reducers/search/filters/locationFilterFunctions.js
mab6bf: In theory, I think we can do that. I had copied this functionality from `bulkAwardTypeChange` in `awardFilterFunctions.js`.

src/js/containers/sharedContainers/LocationListContainer.jsx
mab6bf: I'll also handle this one

src/js/components/search/filters/location/ShownLocation.jsx
```diff
@@ -0,0 +1,51 @@
...
+    location: React.PropTypes.object,
+    removeLocation: React.PropTypes.func
+};
+
+export default class ShownLocation extends React.Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            locationDisplayValue: this.formatLocation()
```
ejgullo: Updated! 👍

src/js/components/search/filters/location/CountryType.jsx
```diff
@@ -6,39 +6,55 @@
...
+        "All",
+        "USA",
+        "Foreign"
+    ]
+};
 
+export default class CountryType extends React.Component {
     render() {
+        const countries =
+            this.props.countries.map((name) => {
```
kevinli-work: This is getting close, but `defaultChecked` is generally a sign that a component is uncontrolled. Here's an example of what a controlled component looks like: [http://codepen.io/anon/pen/JbVdxB](http://codepen.io/anon/pen/JbVdxB)

Basically a component is controlled if I can update its UI by modifying a parent's state via the Redux dev tools instead of clicking. I'm pretty sure this set up actually passes that test, but `defaultChecked` does not force a checked state, which the codepen above does.

ejgullo: Updated 👍

kevinli-work:
Is the API case-sensitive for the `value` geocomplete value? We're passing back capitalized values, but the [API docs indicate they are lower-cased](https://github.com/fedspendingtransparency/usaspending-api/blob/dev/using_the_api.md#body-description-2).

PR #46, opened 2017-01-27
Feature award summary

ejgullo:
- Created Award Summary Page with Summary Bar and Info Bar (DS-647)
- [ ] Code Reviewed


kevinli-work:
The page does not currently appear to handle invalid award IDs and has weird formatting when the award doesn't exist/when the `awards/` API request is in-flight.

screen shot 2017-01-30 at 11 45 20 am

kevinli-work:
Should the Award IDs match?

idmismatch

src/js/containers/award/AwardContainer.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+class AwardContainer extends React.Component {
+
+    constructor(props) {
+        super(props);
+
+        this.getSelectedAward = this.getSelectedAward.bind(this);
+    }
+
+    getSelectedAward() {
```
kevinli-work: As discussed in person, this container should be responsible for fetching the award data on mount (`componentDidMount`) and pass the award data down the child components.

Additionally, this container should monitor for changes to the award ID URL param in order to update when the URL changes.

src/js/containers/award/AwardContainer.jsx
```diff
@@ -0,0 +1,72 @@
...
+                }
+            ]
+        };
+
+        this.selectedAwardRequest = SearchHelper.fetchAward(selectedAwardParams);
+
+        this.selectedAwardRequest.promise
+            .then((res) => {
+                const data = res.data;
+                const awardData = data.results[0];
```
kevinli-work: The container should also be able to handle a condition where there is no award with the given ID.

src/js/containers/award/AwardInfoContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+import SummaryBar from 'components/award/SummaryBar';
+import AwardInfoBar from 'components/award/AwardInfoBar';
+
+import * as awardActions from 'redux/actions/award/awardActions';
+
+const propTypes = {
+    getSelectedAward: React.PropTypes.func
+};
+
+class AwardInfoContainer extends React.Component {
```
kevinli-work: Once the award fetching logic is taken over by the `AwardContainer`, there may not be a need to keep this as a container; it can be downgraded to a component.

src/js/containers/award/AwardInfoContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+    constructor(props) {
+        super(props);
+
+        this.state = {
```
kevinli-work: As discussed in person, this component's state can be simplified. Most of the state appears to be simply string display of data. Additionally, tracking the award/funding agency toggle can be done its own component (`AgencyInfo`).

src/js/containers/award/AwardContainer.jsx
```diff
@@ -0,0 +1,72 @@
...
+            filters: [
+                {
+                    field: "piid",
+                    value: input,
+                    operation: "equals"
+                }
+            ]
+        };
+
+        this.selectedAwardRequest = SearchHelper.fetchAward(selectedAwardParams);
```
kevinli-work: As discussed in person, the filter parameters may be best generated by a SearchOperation model so as to keep query logic centralized.

Additionally, consider switching the API response to being parsed into a standardized Award model object rather than directly storing the JSON output to Redux.

kevinli-work: As discussed in person, the filter parameters may be best generated by a SearchOperation model so as to keep query logic centralized.

Additionally, consider switching the API response to being parsed into a standardized Award model object rather than directly storing the JSON output to Redux.

src/js/components/award/SummaryBar.jsx
```diff
@@ -0,0 +1,93 @@
...
+        let awardId = null;
+        let parentAwardId = null;
+        let status = null;
+        let parentId = null;
+        if (this.props.selectedAward) {
+            const award = this.props.selectedAward;
+            if (award.parent_award) {
+                parentId = award.parent_award;
+            }
+            else if (!award.parent_award && award.type_description !== "Definitive Contract") {
```
kevinli-work: Comparing against the `type` code (listing can be found in `dataMapping/search/awardTypes.js`) is safer than comparing against the `type_description` string as it is less likely to change.

src/js/components/award/RecipientInfo.jsx
```diff
@@ -0,0 +1,93 @@
...
+                    {businessTypeDesc}
+                </div>
+            </li>
+        );
+        return (
+            <div className="recipient-info">
+                <h4>Recipient</h4>
+                { name }
+                <ul className="recipient-information">
+                    { address }
```
kevinli-work: Consider making these smaller components. This render() function is a bit too complex.

src/js/components/award/RecipientInfo.jsx
```diff
@@ -0,0 +1,93 @@
...
+        const name = (
+            <div className="recipient-name">
+                {recipient.recipient_name}
+            </div>);
+        const address = (
+            <li className="recipient-address">
+                <div className="recipient-address title">
+                    Address
+                </div>
+                <div className="recipient-address name">
```
kevinli-work: This section does not handle foreign addresses, which the API returns with different keys. Consider updating the centralized Award model to handle foreign/domestic locations into a single, consolidated data model.

src/js/components/award/AwardInfoBar.jsx
```diff
@@ -0,0 +1,57 @@
...
+    subtierAgency: React.PropTypes.object,
+    officeAgency: React.PropTypes.object,
+    setAwardValues: React.PropTypes.func
+};
+
+export default class AwardInfoBar extends React.Component {
+
+    constructor(props) {
+        super(props);
+        this.state = {
```
kevinli-work: This state doesn't appear to be used.

src/js/components/award/AgencyInfo.jsx
```diff
@@ -0,0 +1,87 @@
...
+    toggleAgency: React.PropTypes.func,
+    agencySet: React.PropTypes.object,
+    agencyTitle: React.PropTypes.string,
+    setAwardValues: React.PropTypes.func,
+    toptierAgency: React.PropTypes.object,
+    subtierAgency: React.PropTypes.object,
+    officeAgency: React.PropTypes.object
+};
+
+export default class AgencyInfo extends React.Component {
```
kevinli-work: Consider moving the state to track the dropdown agency type here instead of the parent container.

kevinli-work:
AwardSummary changes appear to have broken results table. Please keep testing to make sure that changes don't break existing features.

screen shot 2017-02-06 at 1 26 11 pm

src/js/redux/reducers/award/awardReducer.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * awardReducer.js
+ * Created by Emily Gullo 01/23/2017
+ **/
+
+const initialState = [];
```
kevinli-work: Why is this an array? It seems that you are only holding on to one award at a time, and that you have a child key to access that. (Also why do you need the child key if you're only holding the one award?)

src/js/containers/award/AwardContainer.jsx
```diff
@@ -0,0 +1,99 @@
...
+import _ from 'lodash';
+
+import Award from 'components/award/Award';
+
+import * as SearchHelper from 'helpers/searchHelper';
+import * as awardActions from 'redux/actions/award/awardActions';
+import AwardSummary from 'models/results/award/AwardSummary';
+
+const propTypes = {
+    award: React.PropTypes.oneOfType([
```
kevinli-work: See comments about this in the reducer.

src/js/containers/award/AwardContainer.jsx
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    getSelectedAward() {
+        const input = this.props.params.awardId;
+        if (this.selectedAwardRequest) {
+            // A request is currently in-flight, cancel it
+            this.selectedAwardRequest.cancel();
+        }
+        this.selectedAwardRequest = SearchHelper.fetchAward(input);
```
kevinli-work: Can you define this in the `constructor` first so that it is clear this is an expected class property?

src/js/components/award/Award.jsx
```diff
@@ -0,0 +1,44 @@
...
+ **/
+
+import React from 'react';
+import Header from '../sharedComponents/header/Header';
+import Footer from '../sharedComponents/Footer';
+import Error from '../sharedComponents/Error';
+import AwardInfo from './AwardInfo';
+
+const propTypes = {
+    award: React.PropTypes.oneOfType([
```
kevinli-work: See comments in reducer

src/js/components/award/AwardInfoBar.jsx
```diff
@@ -0,0 +1,41 @@
...
+export default class AwardInfoBar extends React.Component {
+
+    render() {
+        let agencies = null;
+        if (this.props.selectedAward) {
+            agencies = (
+                <div className="award-info-bar">
+                    <AgencyInfo
+                        {...this.props}
+                        toggleAgency={this.props.toggleAgency}
```
kevinli-work: Where are these props (`toggleAgency` and `agencyType`) coming from? It looks like these are contained within the child `AgencyInfo` component.

src/js/components/award/RecipientInfo.jsx
```diff
@@ -0,0 +1,51 @@
...
+        const businessTypeDesc = "Not Available";
+
+        const name = (
+            <div className="recipient-name">
+                {recipient.recipient_name}
+            </div>);
+        return (
+            <div className="recipient-info">
+                <h4>Recipient</h4>
+                { name }
```
kevinli-work: Why is this a variable?

src/js/components/award/RecipientInfo.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+const propTypes = {
+    recipient: React.PropTypes.object
+};
+
+export default class RecipientInfo extends React.Component {
+
+    render() {
+        const recipient = this.props.recipient;
+        const recipientDuns = "Not Available";
```
kevinli-work: Do these mean the API currently doesn't return these two fields? If so, is there a story for it?

src/js/components/award/RecipientAddress.jsx
```diff
@@ -0,0 +1,49 @@
...
+        let cityState = "";
+
+        if (city && stateProvince) {
+            cityState = `${city}, ${stateProvince}`;
+        }
+        else if (!city && !stateProvince) {
+            cityState = null;
+        }
+        else {
+            cityState = `${city} ${stateProvince}`;
```
kevinli-work: This seems to be handling if there is either city or state, but won't this return blank spaces (especially leading spaces if there's no city)?

src/js/components/award/RecipientAddress.jsx
```diff
@@ -0,0 +1,49 @@
...
+        }
+        return (
+            <li className={this.props.type}>
+                <div className={`title ${this.props.type}`}>
+                    {this.props.titleValue}
+                </div>
+                <div className="name address">
+                    {recipient.recipient_street}
```
kevinli-work: It seems from the updated model that this will just return a single concatenated string (without space separation) instead of a multi-line street address when there's more than one of data. For example:

```
Main Building123 First AveUnit 45
```

Instead of
```
Main Building
123 First Ave
Unit 45
```

src/js/containers/award/AwardContainer.jsx
```diff
@@ -0,0 +1,99 @@
...
+            noAward: false
+        };
+    }
+
+    componentDidMount() {
+        this.getSelectedAward();
+    }
+
+    componentDidUpdate() {
+        if (this.props.award.selectedAward) {
```
kevinli-work: Changing a URL to an invalid award ID results in an infinite loop.

src/js/components/award/SummaryBar.jsx
```diff
@@ -0,0 +1,84 @@
...
+                        <div className="item-label">Award ID</div>
+                        <div className="item-value">{award.id}</div>
+                    </div>
+                </li>);
+            if (award.type === "D" && parentId === null) {
+                parentAwardId = null;
+            }
+            else {
+                parentAwardId = (
+                    <li>
```
kevinli-work: These can probably be made into their own components.

src/js/models/results/award/AwardSummary.js
```diff
@@ -12,63 +12,141 @@ import GenericRecord from '../GenericRecord';
...
-        id = data[idField];
-        if (!id) {
-            id = '';
+        awardId = data[idField];
+        if (!awardId) {
+            awardId = '';
         }
     }
-    remappedData.id = id;
+    remappedData.awardId = awardId;
```
kevinli-work: Also if you're going to update table, can you add in links to this page in the ID column?

src/js/components/award/SummaryBar.jsx
```diff
@@ -0,0 +1,84 @@
...
+                parentId = null;
+            }
+
+            summaryTitle = (
+                <h1 className="summary-title">{award.type_description} Summary</h1>);
+            awardId = (
+                <li>
+                    <div className="format-item">
+                        <div className="item-label">Award ID</div>
+                        <div className="item-value">{award.id}</div>
```
ejgullo: You're right!  Updating~

src/js/components/award/SummaryBar.jsx
ejgullo: Updating this for several sections with a generic component.

src/js/components/award/AwardInfo.jsx
```diff
@@ -0,0 +1,64 @@
...
+        this.getStatus = this.getStatus.bind(this);
+    }
+
+    componentWillReceiveProps(nextProps) {
+        if (nextProps.selectedAward !== this.props.selectedAward) {
+            this.getStatus(nextProps.selectedAward);
+        }
+    }
+
+    getStatus(award) {
```
ejgullo: I was attempting to parse it the way it comes back from the API, but you're right, that doesn't jive with the redux state.  I believe it worked some of the time because some of the IDs I was testing with had identical start and end dates.

src/js/models/results/award/AwardSummary.js
```diff
@@ -12,63 +12,141 @@ import GenericRecord from '../GenericRecord';
 const recordType = 'award';
 const fields = [
     'id',
+    'awardId',
     'recipient_name',
+    'date_signed',
     'period_of_performance_start_date',
     'period_of_performance_current_end_date',
-    'total_obligation',
```
ejgullo: Mistakes were made.

src/js/components/award/AwardInfo.jsx
ejgullo: I've also moved this down to the `SummaryBar` component where it is relevant - I don't think the parent object needs it.

src/js/models/results/award/AwardSummary.js
ejgullo: Updating the `awardId` property to `award_id` to be in line with the rest of the values as well.

src/js/components/award/AwardInfo.jsx
ejgullo: Also also, the `parentAwardId` if/else logic in the `SummaryBar` component, I don't know if I'm just getting too nitty gritty, but do you think it's too much logic in the render?

src/js/redux/reducers/award/awardReducer.js
ejgullo: Bad copy/pasta and understanding of the reducer setup.  Working on adjusting accordingly.

ejgullo: As per conversation, altering initial state to be an object, rather than an array.

src/js/containers/award/AwardContainer.jsx
ejgullo: As per conversation, changing check from a prop to a state value.

src/js/components/award/Award.jsx
ejgullo: Fixing and noting. :)

src/js/containers/award/AwardContainer.jsx
ejgullo: Can do!

src/js/components/award/AwardInfoBar.jsx
ejgullo: You are correct, missed update in my refactoring, they do not need to be present in this file, nor are they coming from anywhere.

src/js/components/award/RecipientInfo.jsx
ejgullo: Because I am silly.

src/js/components/award/RecipientInfo.jsx
ejgullo: Just got the var names for duns and parent duns, now handling those.  Waiting on a PR from Arman to pass review to have access to business type.

src/js/components/award/RecipientAddress.jsx
ejgullo: Updating to handle those cases

src/js/components/award/RecipientAddress.jsx
ejgullo: I think I have updated this one in a different attempt to handle null values for any of the lines (as many instances are likely to come back without a line 2 or 3.)

kevinli-work:
Can you align the text so they start at the same place?
screen shot 2017-02-08 at 1 58 39 pm

kevinli-work:
Also can you fix the borders so they don't do the weird getting thinner thing?
screen shot 2017-02-08 at 1 59 42 pm

Speaking of borders, can you clean up the top/bottom borders to match the designs (designs on right)?
compare

src/js/components/award/AwardInfo.jsx
```diff
@@ -15,43 +14,13 @@ const propTypes = {
...
-
     render() {
         return (
             <div>
                 <SummaryBar
                     {...this.props}
-                    awardStatus={this.progress}
                     selectedAward={this.props.selectedAward} />
-                <div className="wrapper">
+                <div className="main-content">
```
kevinli-work: Can you add an `id="main-content"` and change this from a `<div>` to `<main>` tag per [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main) and also to match [the search page](https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/components/search/SearchPage.jsx#L164)?

src/js/components/award/SummaryBar.jsx
```diff
@@ -4,21 +4,52 @@
...
+
+        this.getStatus = this.getStatus.bind(this);
+    }
+
+    componentWillMount() {
+        this.getStatus(this.props.selectedAward);
+    }
+
+    componentWillReceiveProps(nextProps) {
+        if (nextProps.selectedAward !== this.props.selectedAward) {
```
kevinli-work: Consider comparing the `selectedAward`'s ID instead of the whole object.

src/js/components/award/SummaryBar.jsx
```diff
@@ -4,21 +4,52 @@
...
+
+    getStatus(award) {
+        const awardStart = moment(award.period_of_performance_start_date, 'MM-DD-YYYY');
+        const awardEnd = moment(award.period_of_performance_current_end_date, 'MM-DD-YYYY');
+        const current = moment();
+        let progress = "";
+        if (current.isBefore(awardStart)) {
+            progress = "Awarded";
+        }
+        else if (current.isAfter(awardEnd)) {
```
kevinli-work: What happens on the last day of an award? Does `isAfter()` track to midnight (that is; during the day of the last award, it counts as _after_ because midnight has passed but the day is not yet over) or does it account for the passage of the full day (isAfter 11:59:59 PM)?

In the first case, this would cause awards to show as Complete during their last day instead of In Progress (unless that is desired behavior).

src/js/components/award/SummaryBar.jsx
```diff
@@ -4,21 +4,52 @@
...
+            progress = "Complete";
+        }
+        else {
+            progress = "In Progress";
+        }
+        this.progress = progress;
```
kevinli-work: Why is the progress string stored as a class variable? This would make more sense in the component state.

src/js/components/award/AgencyInfo.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+        this.state = {
+            agencyType: "awarding"
+        };
+
+        this.toggleAgency = this.toggleAgency.bind(this);
+    }
+
+    toggleAgency(e) {
+        this.setState({
```
kevinli-work: When the page URL changes from one award to another, the toggle stays on the value it was at before (in other words, if I was looking at Funding Agency, and then I change the URL to a different award, that award will load with Funding Agency pre-selected instead of the default Awarding Agency). Consider monitoring for award ID changes and reset the `agencyType` state back to `awarding` to handle this case.

src/js/components/search/table/cells/ResultsTableGenericCell.jsx
```diff
@@ -32,6 +33,14 @@ export default class ResultsTableGenericCell extends React.Component {
             rowClass += ' last-column';
         }
 
+        if (this.props.column === 'award_id') {
```
kevinli-work: The point of this component (per the component name) is to be generic. Instead of adding column-specific logic to the component, I would recommend creating a new component specifically for the ID column. See `ResultsTable` notes for more implementation details.

src/js/components/search/table/ResultsTable.jsx
```diff
@@ -96,7 +96,6 @@ export default class ResultsTable extends React.PureComponent {
         const columns = this.props.columns.map((column, i) => {
```
kevinli-work: The ResultsTable is actually designed to be able to return different cell components for different columns (see note in `ResultsTableGenericCell`). You'll want to add some logic in here to check if the column is the ID column and then return the custom ID cell instead of the generic cell in that case.

src/js/components/award/AwardInfo.jsx
ejgullo: Can do!

src/js/components/award/SummaryBar.jsx
ejgullo: Good point, no need for the rest.

src/js/components/award/SummaryBar.jsx
ejgullo: Moment's `isBefore()`, `isAfter()`, and `isSame()` functions all allow me to pass a variable dictating the level of granularity for the check (for example, i can set it to day to just compare the day, or I can drill down to minutes and seconds, or up to just the year.)  I've spoken to the UX designer and he's let me know that Start and End date should be inclusive of the current date - if the Start Date is 2/8 and the current date is 2/8, then the status should be `In Progress`.  If the End Date is 2/8 and the current date is 2/8, then the status should be `Complete`.  I'll adjust the logic accordingly with a combo of `isSame` and `isBefore` or `isAfter`.

src/js/components/award/SummaryBar.jsx
ejgullo: Was trying to minimize state, perhaps a bit too much!  I'll adjust this value.

src/js/components/award/AgencyInfo.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+        this.state = {
+            agencyType: "awarding"
+        };
+
+        this.toggleAgency = this.toggleAgency.bind(this);
+    }
+
+    componentWillReceiveProps(nextProps) {
+        if (nextProps.selectedAward.id !== this.props.params.awardId) {
```
kevinli-work: This component doesn't have props named `params`. The only prop is `selectedAward`; you'll either need to pass the URL params down or compare `selectedAward.id` values.

src/js/components/award/SummaryBar.jsx
```diff
@@ -0,0 +1,107 @@
...
+            this.getStatus(nextProps.selectedAward);
+        }
+    }
+
+    getStatus(award) {
+        const awardStart = moment(award.period_of_performance_start_date, 'MM-DD-YYYY');
+        const awardEnd = moment(award.period_of_performance_current_end_date, 'MM-DD-YYYY');
+        const current = moment();
+        let progress = "";
+        if (current.isBefore(awardStart, 'day') || current.isSame(awardStart, 'day')) {
```
kevinli-work: [http://momentjs.com/docs/#/query/is-same-or-before/](http://momentjs.com/docs/#/query/is-same-or-before/)

[http://momentjs.com/docs/#/query/is-same-or-after/](http://momentjs.com/docs/#/query/is-same-or-after/)

src/js/components/award/AgencyInfo.jsx
kevinli-work: Idk what happens in the latter case if the second URL is an invalid ID.

PR #48, opened 2017-02-01
Promote dev-stable to staging

kevinli-work:
* Sprint 31


afrasier:
whoops this isn't even my jam, my bad @willkjackson

PR #53, opened 2017-02-08
Autocomplete Replacement

mab6bf:
This has been implemented in the Location Filter only. I will add to Agency Filter after it is PRd, also removing the Typeahead.jsx component and the inclusion of awesomplete in package.json.


src/js/containers/search/filters/location/LocationListContainer.jsx
```diff
@@ -99,13 +100,33 @@ class LocationListContainer extends React.Component {
     }
 
     render() {
+        const values = [];
```
kevinli-work: Can we move the parsing of autocomplete locations out of the `render()` function, which may run multiple times for reasons other than autocomplete changes? This prevents the loop from having to needlessly run; consider putting this in state instead.

src/js/components/sharedComponents/autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,242 @@
...
+
+    componentDidUpdate(prevProps) {
+        if (!_.isEqual(prevProps.values, this.props.values)) {
+            this.open();
+        }
+    }
+
+    onChange(e) {
+        this.checkValidity(e.target.value);
+        this.props.handleTextInput(e);
```
kevinli-work: Should `state.selectedIndex` reset to zero when the autocomplete results change? It currently doesn't, so if you have second item highlighted and then change the search term, the second item of the new list is pre-selected.

src/js/components/sharedComponents/autocomplete/Suggestion.jsx
```diff
@@ -0,0 +1,50 @@
...
+    data: []
+};
+
+export default class Suggestion extends React.Component {
+    componentDidMount() {
+        this.setUpSuggestion();
+    }
+
+    setUpSuggestion() {
+        this.suggestion.addEventListener('mousedown', () => {
```
kevinli-work: Why not nest a `button` inside the list item, to make it aria-friendly?

src/js/components/sharedComponents/autocomplete/SuggestionHolder.jsx
```diff
@@ -0,0 +1,59 @@
...
+        this.hidden = true;
+    }
+
+    render() {
+        const suggestions = [];
+        if (this.props.suggestions.length > 0 && this.props.shown !== 'hidden') {
+            this.hidden = false;
+        }
+        else {
+            this.hidden = true;
```
kevinli-work: Why is `this.hidden` a class variable instead of state?

src/js/components/sharedComponents/autocomplete/SuggestionHolder.jsx
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        // Ensure we're only showing 10 results at most
+        for (let i = 0; i < Math.min(this.props.suggestions.length, 10); i++) {
```
kevinli-work: Might make sense to make the max suggestion length a prop (with `defaultProps` set to 10), so autocomplete is more easily customized in the future.

src/js/components/sharedComponents/autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,242 @@
...
+    previous() {
+        if (this.state.selectedIndex > 0) {
+            this.setState({
+                selectedIndex: this.state.selectedIndex -= 1
+            });
+        }
+    }
+
+    next() {
+        if (this.state.selectedIndex < 9) {
```
kevinli-work: If you make the suggestion list length a prop (see comment in `SuggestionHolder`), you'll also want to modify this number.

src/js/components/sharedComponents/autocomplete/Autocomplete.jsx
kevinli-work: This will also prevent keyboard weirdness when an item is highlighted/selected on a long list, and then the input changes the suggestion list to something that is shorter than the previous selected index.

mab6bf: Yep, good call

src/js/components/sharedComponents/autocomplete/SuggestionHolder.jsx
mab6bf: Done

src/js/containers/search/filters/location/LocationListContainer.jsx
mab6bf: Done

src/js/components/sharedComponents/autocomplete/SuggestionHolder.jsx
mab6bf: Done

src/js/components/sharedComponents/autocomplete/Autocomplete.jsx
mab6bf: Done

PR #60, opened 2017-02-21
Recipient Filter

mab6bf:


src/js/containers/search/filters/recipient/RecipientLocationContainer.jsx
```diff
@@ -0,0 +1,156 @@
...
+            });
+
+            if (this.recipientLocationSearchRequest) {
+                // A request is currently in-flight, cancel it
+                this.recipientLocationSearchRequest.cancel();
+            }
+
+            const locSearchParams = {
+                value: this.state.recipientLocationSearchString,
+                scope: this.props.locationDomesticForeign,
```
kevinli-work: Scope is not working, it appears this prop should be named `recipientDomesticForeign`.

src/js/containers/search/filters/recipient/RecipientNameDUNSContainer.jsx
```diff
@@ -0,0 +1,147 @@
...
+                fields: ['recipient_name', 'recipient_unique_id'],
+                value: this.state.recipientSearchString,
+                matched_objects: true
+            };
+
+            this.recipientSearchRequest = SearchHelper.fetchRecipients(recipientSearchParams);
+
+            this.recipientSearchRequest.promise
+                .then((res) => {
+                    const data = res.data.matched_objects.recipient_name;
```
kevinli-work: This only returns recipients whose names matched the autocomplete string. DUNS matches are not included because the API puts those results in `recipient_unique_id`. Also, don't forget to remove duplicates that be returned by both fields.

src/js/components/search/filters/recipient/SelectedRecipientLocations.jsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export default class SelectedRecipientLocations extends React.Component {
+    constructor(props) {
+        super(props);
+
+        this.formatLocation = this.formatLocation.bind(this);
+    }
+
+    formatLocation(loc) {
```
kevinli-work: Let's consider moving the location tag formatting code to a central helper.

src/js/containers/search/filters/recipient/RecipientLocationContainer.jsx
```diff
@@ -0,0 +1,156 @@
...
+                usage: "location"
```
kevinli-work: API docs say the value for this field should be `recipient`. Can you confirm?

src/js/containers/search/filters/recipient/RecipientLocationContainer.jsx
mab6bf: Fixed

src/js/containers/search/filters/recipient/RecipientLocationContainer.jsx
mab6bf: It should indeed be `recipient` - fixed

src/js/containers/search/filters/recipient/RecipientNameDUNSContainer.jsx
mab6bf: Great catch - fixed

src/js/components/search/filters/recipient/SelectedRecipientLocations.jsx
mab6bf: Done - implemented in Location Filter as well

PR #62, opened 2017-02-23
Top Bar Implementation for Agency Filter

mab6bf:


mab6bf:
FYI, this won't pass Travis until the .eslintrc changes from #60 are merged in

src/js/containers/search/topFilterBar/TopFilterBarContainer.jsx
```diff
@@ -176,6 +186,44 @@ export class TopFilterBarContainer extends React.Component {
     }
 
     /**
+     * Logic for parsing the current Redux selected Awarding and Funding Agencies into a JS object
+     * that can be parsed by the top filter bar
+     */
+    prepareAgencies(props, type) {
```
kevinli-work: Maybe we should pass an additional argument in here to specify removeFilter operation, so we don't have the increasing blob that is 
```
else if (type === 'selectedLocations' || type === 'selectedFundingAgencies' || type === 'selectedAwardingAgencies')
```
around line 257, or alternatively expose the props to the TopFilterGroups so the individual filters can choose the appropriate removal function.

PR #74, opened 2017-02-28
Add “coming-soon” labels to unfinished features on the Search page

AGuyNamedMarco:
Add "coming soon" labels for the following features:

**Search Results Page**
- Spending over time graph, months and quarters options
- free text search filter
- Budget categories filters
- Award id filter
- Award amount filter
- contract specific details filter
- Direct payments, loans, and insurance tabs on Award table
- Offices option on "Spending By" Graph

### DIFF
--- a/src/_scss/pages/search/_searchOption.scss
+++ b/src/_scss/pages/search/_searchOption.scss
@@ -17,5 +17,46 @@
 
     }
 
+    &.coming-soon {
+        background: repeating-linear-gradient(
+            45deg,
+            rgba($color-gray-lightest, .2),
+            rgba($color-gray-lightest, .2) rem(2px),
+            rgba($color-gray-lighter, .3) rem(2px),
+            rgba($color-gray-lighter, .3) rem(4px)
+        );
+        .coming-soon-icon {
+            bottom: rem(24px);
+            height: 0;
+            left: rem(35px);
+            position: relative;
+            svg {
+                fill: $color-gray-lighter;
+                height: rem(10px);
+                width: rem(10px);
+            }
+        }
+        .coming-soon-label {
+            bottom: rem(20px);
+            color: $color-gray-lighter;
+            float: left;
+            font-weight: bold;
+            font-size: rem(10px);
+            height: 0;
+            left: rem(50px);
+            position: relative;
+        }
+        .filter-toggle {
+            svg {
+                fill: $color-gray-lighter;
+            }
+            .filter-header {
+                color: $color-gray-lighter;
+            }
+        }
+        &:hover * {
+            cursor: default;
+        }
+    }
     @import "./_filterExpand";
 }
--- a/src/_scss/pages/search/results/table/_tableTypes.scss
+++ b/src/_scss/pages/search/results/table/_tableTypes.scss
@@ -29,5 +29,53 @@
             border-left: 1px solid $color-gray-light;
             color: $color-base;
         }
+        &.coming-soon {
+            background: repeating-linear-gradient(
+                45deg,
+                rgba($color-gray-lightest, .2),
+                rgba($color-gray-lightest, .2) rem(2px),
+                rgba($color-gray-lighter, .3) rem(2px),
+                rgba($color-gray-lighter, .3) rem(4px)
+            );
+            color: $color-gray-light;
+            .coming-soon-container {
+                width: 100%;
+                text-align: center;
+                height: 10px;
+            }
+            .coming-soon-icon {
+                bottom: rem(0px);
+                display: inline;
+                left: rem(0px);
+                svg {
+                    fill: $color-gray-lighter;
+                    height: rem(10px);
+                    width: rem(10px);
+                }
+            }
+            .coming-soon-label {
+                bottom: rem(2px);
+                color: $color-gray-lighter;
+                display: inline;
+                font-weight: bold;
+                font-size: rem(10px);
+                height: 0;
+                left: rem(3px);
+                position: relative;
+            }
+            &:hover {
+                background: repeating-linear-gradient(
+                    45deg,
+                    rgba($color-gray-lightest, .2),
+                    rgba($color-gray-lightest, .2) rem(2px),
+                    rgba($color-gray-lighter, .3) rem(2px),
+                    rgba($color-gray-lighter, .3) rem(4px)
+                );
+                cursor: default;
+            }
+            &:hover * {
+                cursor: default;
+            }
+        }
     }
 }
--- a/src/_scss/pages/search/results/visualizations/rank/_topSection.scss
+++ b/src/_scss/pages/search/results/visualizations/rank/_topSection.scss
@@ -19,6 +19,10 @@
     @import "../../../../../components/visualizations/_scope";
 
     .content {
+        @include media($tablet-screen) {
+            width: 300px;
+            height: 60px;
+        }
         @include media($large-screen) {
             width: 390px;
             padding: 15px 30px;
--- a/src/_scss/pages/search/results/visualizations/rank/_topSection.scss
+++ b/src/_scss/pages/search/results/visualizations/rank/_topSection.scss
@@ -37,4 +37,41 @@
             }
         }
     }
+    .coming-soon {
+        .scope-button {
+            color: $color-gray-lighter;
+        }
+        .scope-button:hover {
+            color: $color-gray-lighter;
+            border-bottom: 4px solid rgba(255,255,255,0);
+        }
+        .coming-soon-icon {
+            bottom: rem(10px);
+            height: 0;
+            float: left;
+            left: rem(0px);
+            position: relative;
+            svg {
+                fill: $color-gray-lighter;
+                height: rem(10px);
+                width: rem(10px);
+            }
+        }
+        .coming-soon-label {
+            bottom: rem(6px);
+            color: $color-gray-lighter;
+            float: left;
+            font-weight: bold;
+            font-size: rem(10px);
+            height: 0;
+            left: rem(13px);
+            position: relative;
+        }
+        &:hover {
+            cursor: default;
+        }
+        &:hover * {
+            cursor: default;
+        }
+    }
 }
--- a/src/_scss/pages/search/results/visualizations/time/_topSection.scss
+++ b/src/_scss/pages/search/results/visualizations/time/_topSection.scss
@@ -22,13 +22,11 @@
         @include media($tablet-screen) {
             width: 300px;
             height: 60px;
-
-            @include media($large-screen) {
-                width: 390px;
-                padding: 15px 30px;
-            }
         }
-
+        @include media($large-screen) {
+            width: 390px;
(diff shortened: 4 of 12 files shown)


src/js/components/search/FilterExpandButton.jsx
```diff
@@ -25,10 +25,32 @@ export default class FilterExpandButton extends React.Component {
...
+                break;
+            case 'Award Amount':
+                clickEvent = null;
+                break;
+            case 'Other Award Items':
+                clickEvent = null;
+                break;
+            default:
+                clickEvent = this.props.toggleFilter;
+                break;
```
ejgullo: Since clickEvent is null in all but the default case, you might consider letting your initial declaration be null, and only declare it again in the default case.  If it's initially set to null, you won't need to re-set it unless it is something other than null.

src/js/components/search/SearchOption.jsx
```diff
@@ -47,41 +48,63 @@ export default class SearchOption extends React.Component {
...
+                comingSoon = '';
                 break;
             case 'Recipients':
                 searchOption = (<RecipientSearchContainer />);
+                statusClass = '';
+                comingSoon = '';
                 break;
             default:
-                searchOption = null;
+                searchOption = comingSoon;
```
ejgullo: It may be cleaner to let `comingSoon` be null by default, and let it be the declared value in the default case.  (I'll defer to Kevin on this.)

src/_scss/pages/search/_searchOption.scss
```diff
@@ -17,5 +17,46 @@
...
+        }
+        .filter-toggle {
+            svg {
+                fill: $color-gray-lighter;
+            }
+            .filter-header {
+                color: $color-gray-lighter;
+            }
+        }
+        &:hover * {
```
ejgullo: Is the wildcard selector necessary here?

src/_scss/pages/search/results/table/_tableTypes.scss
```diff
@@ -29,5 +29,53 @@
...
+                background: repeating-linear-gradient(
+                    45deg,
+                    rgba($color-gray-lightest, .2),
+                    rgba($color-gray-lightest, .2) rem(2px),
+                    rgba($color-gray-lighter, .3) rem(2px),
+                    rgba($color-gray-lighter, .3) rem(4px)
+                );
+                cursor: default;
+            }
+            &:hover * {
```
ejgullo: Same as above - also I think in this case, the wildcard hover is just duplicating the above hover class.

src/_scss/pages/search/results/visualizations/rank/_topSection.scss
```diff
@@ -37,4 +37,41 @@
...
+            font-weight: bold;
+            font-size: rem(10px);
+            height: 0;
+            left: rem(13px);
+            position: relative;
+        }
+        &:hover {
+            cursor: default;
+        }
+        &:hover * {
```
ejgullo: Same as above.

src/_scss/pages/search/results/visualizations/time/_topSection.scss
```diff
@@ -40,5 +40,42 @@
...
+                font-weight: bold;
+                font-size: rem(10px);
+                height: 0;
+                left: rem(13px);
+                position: relative;
+            }
+            &:hover {
+                cursor: default;
+            }
+            &:hover * {
```
ejgullo: Same as above.

src/js/components/search/SearchOption.jsx
```diff
@@ -47,41 +48,63 @@ export default class SearchOption extends React.Component {
...
+                statusClass = 'coming-soon';
         }
 
         if (this.state.showFilter !== true) {
             searchOption = null;
         }
 
         return (
-            <div className="search-option">
+            <div className={'search-option ' + statusClass}>
```
ejgullo: Does this className reflect as expected?  If not, look at a similar mixed text + variable className in `ResultsTableTabItem.jsx`

src/js/components/search/table/ResultsTableTabItem.jsx
```diff
@@ -24,17 +25,38 @@ export default class ResultsTableTabItem extends React.Component {
     }
 
     render() {
-        let activeClass = '';
+        let activeClass = '',
+            comingSoon = '',
+            status = '',
+            clickEvent = '';
```
ejgullo: Make sure to declare each item with `let` individually.

src/js/components/search/table/ResultsTableTabItem.jsx
```diff
@@ -24,17 +25,38 @@ export default class ResultsTableTabItem extends React.Component {
...
+            comingSoon = '';
+            status = '';
+            clickEvent = this.clickedTab
+        }
 
         return (
             <button
-                className={`table-type-toggle${activeClass}`}
-                onClick={this.clickedTab}
+                className={`table-type-toggle${activeClass}` + ' ' + status}
```
ejgullo: As with above, I'm not certain this concatenation style works with react - if it's not working as you expect, check another similar concat style (`ResultsTable.jsx`)

src/js/components/search/visualizations/rank/RankVisualizationScopeButton.jsx
```diff
@@ -34,13 +34,20 @@ export default class RankVisualizationScopeButton extends React.Component {
             description += ' (currently selected)';
         }
 
+        let clickEvent = '';
+        if (this.props.label == 'Offices') {
```
ejgullo: In react, we need to use triple comparison operators - `===`, `!==` .

src/js/components/search/visualizations/rank/RankVisualizationScopeButton.jsx
```diff
@@ -34,13 +34,20 @@ export default class RankVisualizationScopeButton extends React.Component {
...
+        }
+
         return (
             <button
                 className={`scope-button${activeClass}`}
                 value={this.props.value}
                 title={description}
                 aria-label={description}
-                onClick={this.clickedButton}>
+                onClick={clickEvent}>
```
kevinli-work: Instead of variably nulling the `onClick` event, you should set the `disabled` attribute to `true`. This prevents the click event from occurring and also informs screen readers for visually impaired users that the button is inoperable.

src/js/components/search/table/ResultsTableTabItem.jsx
kevinli-work: The tick syntax is an [ES6 template string](https://developers.google.com/web/updates/2015/01/ES6-Template-Strings), which means you can do concatenation without `+`:

```
const myName = 'Kevin';

const concatString = `Hello ${myName}`;
```
outputs `Hello Kevin`

src/js/components/search/FilterExpandButton.jsx
kevinli-work: 1) See comment in RankVisualizationScopeButton about the `disabled` attribute to disable a button
2) The FilterExpandButton component should not care what filter it belongs to. The parent component should pass a prop to it indicating its enable/disabled state, which in turn sets the `<button>` element's `disabled` attribute

src/js/components/search/SearchOption.jsx
kevinli-work: I agree. Set it to null in line 52, set it to the JSX component in the default block in line 90.

src/js/components/search/SearchOption.jsx
```diff
@@ -47,62 +47,59 @@ export default class SearchOption extends React.Component {
     }
 
     render() {
+        let disabledStatus = false;
+        let comingSoon = null;
+        let comingSoonModule = (
```
kevinli-work: This needs to be `const` to pass the linting (getting a green checkmark in Github instead of the red X)

src/js/components/search/FilterExpandButton.jsx
```diff
@@ -25,32 +25,20 @@ export default class FilterExpandButton extends React.Component {
...
-            default:
-                clickEvent = this.props.toggleFilter;
-                break;
+        let disabledStatus = false;
+        if (this.props.name !== 'Search' ||
+            this.props.name !== 'Budget Categories' ||
+            this.props.name !== 'Award ID' ||
+            this.props.name !== 'Award Amount' ||
+            this.props.name !== 'Other Award Items') {
+                disabledStatus = true;
```
kevinli-work: This needs to un-indented by 4 spaces (one tab to the left) to pass the linting (getting a green checkmark in Github instead of the red X).

src/js/components/search/table/ResultsTableTabItem.jsx
```diff
@@ -25,36 +26,36 @@ export default class ResultsTableTabItem extends React.Component {
     }
 
     render() {
-        let activeClass = '',
-            comingSoon = '',
-            status = '',
-            clickEvent = '';
+        let activeClass = '';
+        let comingSoon = '';
+        let comingSoonModule = (
```
kevinli-work: This needs to be `const` to pass the linting (getting a green checkmark in Github instead of the red X)

src/js/components/search/visualizations/rank/RankVisualizationScopeButton.jsx
```diff
@@ -34,20 +34,14 @@ export default class RankVisualizationScopeButton extends React.Component {
...
-
         return (
             <button
                 className={`scope-button${activeClass}`}
                 value={this.props.value}
                 title={description}
                 aria-label={description}
-                onClick={clickEvent}>
+                onClick={this.clickedButton}
+                disabled={this.props.label === 'Offices' ? true : false}>
```
kevinli-work: Overcomplex ternary operations are [not allowed](http://eslint.org/docs/rules/no-unneeded-ternary) by our linter. You can switch this line to a simpler version:

```
disabled={this.props.label !== 'Offices'}>
```

src/_scss/pages/search/results/table/_tableTypes.scss
AGuyNamedMarco: See my response in the "_searchOption.scss" thread

src/_scss/pages/search/_searchOption.scss
AGuyNamedMarco: The wildcard was needed (in addition to the bare :hover pseudo-class) in order to apply the "cursor: default" style to the containing element as well as the children (the arrow icon, link text, and "coming-soon" label.

src/_scss/pages/search/results/visualizations/rank/_topSection.scss
AGuyNamedMarco: See my response in the "_searchOption.scss" thread

src/_scss/pages/search/results/visualizations/time/_topSection.scss
AGuyNamedMarco: See my response in the "_searchOption.scss" thread

PR #86, opened 2017-03-07
Award ID Filter

mab6bf:


kevinli-work:
Testing:

screen shot 2017-03-07 at 5 07 13 pm
* Unknown award error message pops up even when autocomplete returns results

screen shot 2017-03-07 at 5 08 12 pm
* Need to `word-wrap: break-word;` on these long IDs
* Arrow key weirdness, I can apparently keep highlighting nonexistent rows; to get back to the first result I have to press the up key a bunch of times

src/js/containers/search/filters/awardID/AwardIDListContainer.jsx
```diff
@@ -0,0 +1,209 @@
...
+
+        Object.keys(fieldMap).forEach((key) => {
+            const awardIDs = awardIDProps[key];
+            if (awardIDs && awardIDs.length > 0) {
+                awardIDs.forEach((item) => {
+                    if (item) {
+                        values.push({
+                            title: AwardIDFormatter.formatAwardID(item, fieldMap[key]),
+                            subtitle: fieldMap[key],
+                            data: Object.assign({}, item, {
```
kevinli-work: Do we need to retain the entire `item` or can we just grab the `id`? Especially as these objects get bigger and we're holding increasingly large amounts of data in state...

src/js/containers/search/filters/awardID/AwardIDListContainer.jsx
```diff
@@ -0,0 +1,209 @@
...
+                value: this.state.awardIDSearchString,
+                matched_objects: true
+            };
+
+            this.awardIDSearchRequest = SearchHelper.fetchAwardIDs(awardIDSearchParams);
+
+            this.awardIDSearchRequest.promise
+                .then((res) => {
+                    this.setState({
+                        noResults: !res.data.matched_objects.length
```
kevinli-work: I don't think this is correctly setting the flag. Maybe an explicit `res.data.matched_objects.length === 0`?

src/js/containers/search/filters/awardID/AwardIDListContainer.jsx
```diff
@@ -0,0 +1,209 @@
...
+                    this.props.selectedAwardIDs.toArray().forEach((item) => {
+                        selectedAwardIDs.push(_.omit(item, 'awardIDType'));
+                    });
+
+                    // Filter out any selectedAwardIDs that may be in the result set
+                    if (selectedAwardIDs && selectedAwardIDs.length > 0) {
+                        // We need to compare selectedAwardIDs to each result object,
+                        // as we're receiving a result object with 4 keys:
+                        // PIID, FAIN, URI, and PARENT_AWARD__PIID
+                        Object.keys(data).forEach((key) => {
```
kevinli-work: Why not just check the `id` value instead of the whole object? Isn't it supposed to be the DB primary key and thus unique?

src/js/containers/search/filters/awardID/AwardIDListContainer.jsx
```diff
@@ -0,0 +1,209 @@
...
+                            autocompleteData[key] =
+                                _.differenceWith(data[key], selectedAwardIDs, _.isEqual);
+                        });
+                    }
+                    else {
+                        autocompleteData = data;
+                    }
+
+                    this.noResults = !autocompleteData.length;
```
kevinli-work: This looks like old code.

src/js/containers/search/filters/awardID/AwardIDSearchContainer.jsx
```diff
@@ -0,0 +1,47 @@
...
+export class AwardIDSearchContainer extends React.Component {
+    constructor(props) {
+        super(props);
+
+        // Bind function
+        this.toggleAwardID = this.toggleAwardID.bind(this);
+    }
+
+    toggleAwardID(awardID) {
+        const updateParams = {};
```
mab6bf: Done

src/js/containers/search/filters/awardID/AwardIDListContainer.jsx
```diff
@@ -0,0 +1,209 @@
...
+        this.timeout = null;
+        this.fields = ['piid', 'fain', 'uri', 'parent_award__piid'];
+    }
+
+    componentDidMount() {
+        this.parseAutocompleteAwardIDs(this.props);
+    }
+
+    componentWillReceiveProps(nextProps) {
+        if (!_.isEqual(nextProps.piid, this.props.piid)
```
mab6bf: Good call - fixed

src/js/containers/search/filters/awardID/AwardIDListContainer.jsx
mab6bf: Correct - fixed. I think this was causing the error in the above screenshot.

src/js/containers/search/filters/awardID/AwardIDListContainer.jsx
mab6bf: Indeed - fixed to setState instead

src/js/containers/search/topFilterBar/TopFilterBarContainer.jsx
```diff
@@ -328,7 +357,7 @@ export class TopFilterBarContainer extends React.Component {
         }
         else if (type === 'selectedLocations' || type === 'selectedFundingAgencies'
```
mab6bf: Converted to a switch

src/js/containers/search/filters/awardID/AwardIDListContainer.jsx
mab6bf: Created a new object with `id`, `piid`, `fain`, `uri`, and the internal tracking of which category item it is

src/js/containers/search/filters/awardID/AwardIDListContainer.jsx
mab6bf: Done

PR #87, opened 2017-03-07
Feature award contract

ejgullo:
- Tests written for Contract Summary
- Passing down inflight to details table to provide loading/no results messaging
- Moved things into functions and out of render
- Refactored graph visualization


tests/containers/award/AwardContainer-test.jsx
```diff
@@ -0,0 +1,475 @@
...
+                setAwardTransactions={AwardActions.setAwardTransactions}
+                appendAwardTransactions={AwardActions.appendAwardTransactions}
+                setTransactionsMeta={AwardActions.setTransactionsMeta}
+                updateTransactionRenderHash={AwardActions.updateTransactionRenderHash} />);
+
+        const parseTxnSpy = sinon.spy(awardContainer.instance(),
+            'parseTransactions');
+
+        // Parse award
+        awardContainer.instance().parseTransactions(txnData);
```
kevinli-work: Same as above; you need to actually test that `parseTransactions` is correctly parsing.

tests/containers/award/AwardContainer-test.jsx
```diff
@@ -0,0 +1,475 @@
...
+                award={award} />);
+
+        const getNextPageSpy = sinon.spy(awardContainer.instance(),
+            'nextTransactionPage');
+
+        // Retrieve award and transactions
+        awardContainer.instance().nextTransactionPage();
+
+        // checking that it ran
+        expect(getNextPageSpy.callCount).toEqual(1);
```
kevinli-work: We need to test more than just that this function is called. We need to validate that `nextTransactionPage` actually increases the pagination.

tests/redux/reducers/award/awardReducer-test.jsx
```diff
@@ -0,0 +1,420 @@
...
+                        action_type: null,
+                        federal_action_obligation: '-$14,274',
+                        modification_number: '860',
+                        description: 'MANAGEMENT AND OPERATION OF THE OAK RIDGE NATIONAL LABORATORY - RECOVERY TAS::89 0227::TAS - DEOBLIGATE SC ARRA FUNDING FROM PROJECT ENTITLED "NUCLEAR PHYSICS - NUCLEAR SCIENCE WORKFORCE"'
+                    }
+                ]
+            };
+            const updatedState = awardReducer(initialState, action);
+            // the value should be equal
+            expect(updatedState.transactions).toEqual(action.transactions);
```
kevinli-work: The initial state is an empty array, so appending data to an empty array doesn't give us assurance that this Redux action is actually appending and not replacing.

tests/redux/reducers/award/awardReducer-test.jsx
```diff
@@ -0,0 +1,420 @@
...
+
+    describe('UPDATE_TXN_RENDER_HASH', () => {
+        it('should provide a unique hash id for the transaction render', () => {
+            const action = {
+                type: 'UPDATE_TXN_RENDER_HASH',
+                renderHash: _.uniqueId()
+            };
+            const updatedState = awardReducer(initialState, action);
+            // the value should be equal
+            expect(updatedState.renderHash).not.toEqual(action.renderHash);
```
kevinli-work: To validate that the hash has changed, we need to execute the Redux function multiple times (twice) and test that the hash changed between those times.

tests/redux/reducers/award/awardReducer-test.jsx
```diff
@@ -0,0 +1,420 @@
...
+
+    describe('UPDATE_TXN_GROUP_HASH', () => {
+        it('should provide a unique hash for the transaction group', () => {
+            const action = {
+                type: 'UPDATE_TXN_GROUP_HASH',
+                groupHash: _.uniqueId()
+            };
+            const updatedState = awardReducer(initialState, action);
+            // the value should be equal
+            expect(updatedState.groupHash).not.toEqual(action.groupHash);
```
kevinli-work: Same as above

tests/redux/reducers/award/awardReducer-test.jsx
```diff
@@ -0,0 +1,420 @@
...
+    });
+
+    describe('RESET_AWARD_DATA', () => {
+        it('should reset state to the initial state', () => {
+            const action = {
+                type: 'RESET_AWARD_DATA'
+            };
+            const updatedState = awardReducer(initialState, action);
+            // the value should be equal
+            expect(updatedState).toEqual(initialState);
```
kevinli-work: This doesn't give us assurance that the reset Redux action works, since we are simply resetting the initial state back to the initial state...which means nothing is happening. You need to modify the Redux state to be something different from the initial state, perform the reset Redux action, and then test that the new Redux state matches the initial state.

src/js/components/award/ContractDetails.jsx
```diff
@@ -13,25 +13,53 @@ const propTypes = {
...
+            typeDesc: "",
+            price: ""
+        };
+
+        // bind functions
+        this.setValues = this.setValues.bind(this);
+    }
+
+    componentDidMount() {
+        this.setValues(this.props.selectedAward);
```
ejgullo: Noted!  I'll update accordingly

src/js/components/award/visualizations/AwardLabelsLine.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+const propTypes = {
+    labelDistance: React.PropTypes.number,
+    type: React.PropTypes.string,
+    path: React.PropTypes.string
+};
+
+export default class AwardLabelsLine extends React.Component {
+
+    render() {
```
ejgullo: Will do!

src/js/components/award/visualizations/AmountsChart.jsx
```diff
@@ -170,89 +170,26 @@ export default class AmountsChart extends React.Component {
                     }}>
 
                     <g transform="translate(0, 5)">
```
ejgullo: Moving the polyline generation down to the (new) polyline component!

ejgullo: @kevinli-work Would it be worthwhile to move the drawing of the bars themselves down to their individual components as well?

kevinli-work: It depends on the situation. If you can, it would be best, but if the bars are heavily dependent on other components or their drawing calculation output needs to be used by other components, it may be better to keep it in the parent component.

tests/containers/award/AwardContainer-test.jsx
```diff
@@ -0,0 +1,475 @@
...
+            'parseAward');
+
+        // Parse award
+        awardContainer.instance().parseAward(fullData);
+
+        // checking that it ran
+        expect(parseAwardSpy.callCount).toEqual(1);
+
+        // reset the spies
+        parseAwardSpy.reset();
```
ejgullo: 👍

tests/containers/award/AwardContainer-test.jsx
ejgullo: 👍

ejgullo: @kevinli-work I noticed in this one specifically, the test does not like it if other parts of the function are not explicitly defined (`updateTransactionRenderHash` for example, passed into the mocked `AwardContainer`).  I'm assuming my current method of declaring them as `AwardAction.updateTransactionRenderHash` is a bad idea, and I should be using mocked actions.  But as those functions are not being used in this particular test, is it necessary?

kevinli-work: `updateTransactionRenderHash` is a Redux action (passed in as a prop function) that is used inside `parseTransactions`. The same goes for all the other `this.props.xxxxx()` functions that are inside `parseTransactions`, as a result they are needed.

Remember that when testing, you don't actually have Redux, so none of these Redux actions are available or automatically connected via props. Even if these actions are available (if you `import` them), they won't do anything because there's no Redux store to act on. Instead you need to manually create props for each function when you shallow mount the container, and these props should be mocked functions.

kevinli-work: See [https://github.com/fedspendingtransparency/usaspending-website/blob/dev/tests/containers/search/recipientFilter/RecipientSearchContainer-test.jsx#L532](https://github.com/fedspendingtransparency/usaspending-website/blob/dev/tests/containers/search/recipientFilter/RecipientSearchContainer-test.jsx#L532)

tests/containers/award/AwardContainer-test.jsx
ejgullo: 👍

tests/redux/reducers/award/awardReducer-test.jsx
ejgullo: 👍 Will be setting a separate `appendState` that contains a single transaction, and check that the `updatedState` value equals a concat of that `appendState` and the `action` items.

tests/redux/reducers/award/awardReducer-test.jsx
ejgullo: 👍

tests/redux/reducers/award/awardReducer-test.jsx
ejgullo: 👍

tests/redux/reducers/award/awardReducer-test.jsx
ejgullo: 👍

tests/containers/award/AwardContainer-test.jsx
```diff
@@ -365,105 +367,231 @@ const txnData = {
...
+        expect(mockReduxAward).toHaveBeenCalledTimes(1);
 
         // reset the spies
         parseAwardSpy.reset();
     });
 
     // parse transactions
     it('should parse the returned transactions and send to redux store', () => {
+        const expectedTxnData = [{
+            _jsid: "contract-transaction-3",
```
kevinli-work: `_jsid` is a unique value which is controlled by the lodash library, so I'm not sure we can be certain what to expect from it. To avoid the test randomly breaking, I'd suggest excluding `_jsid` from the expected value and deleting the `_jsid` property from `args`  (`delete args._jsid`).

tests/containers/award/AwardContainer-test.jsx
```diff
@@ -365,105 +367,231 @@ const txnData = {
...
+        const mockAppendAwardTxn = jest.fn((args) => {
+            expect(args).toEqual(expectedTxnData);
+        });
+
+        const mockSetTxnMeta = jest.fn((args) => {
+            expect(args).toEqual(award.transactionMeta);
+        });
+
+        const mockUpdateTxnRenderHash = jest.fn((args) => {
+            expect(args).toEqual(undefined);
```
kevinli-work: Since these two functions don't take arguments, you can do an empty mock for them (`jest.fn();`) and then just check they they were called:
```
const mockUpdateTxnRenderHash = jest.fn();

expect(mockUpdateTxnRenderHash).toHaveBeenCalled();
```

tests/containers/award/AwardContainer-test.jsx
```diff
@@ -365,105 +367,231 @@ const txnData = {
...
     });
 
     // next transaction page
     it('should fetch the next page of available transactions', () => {
+        // set initial page
+        const initialTxnMeta = {
+            page: 1
+        };
+        // Mock the call and validate it does NOT the expected award page
+        const MockReduxAwardTxnMeta = jest.fn((args) => {
```
kevinli-work: Lowercase the `M` please.

Also I'm still not sure this test is actually testing anything. `nextTransactionPage()` is not supposed to do anything if we're currently on the last page. The `award` Redux value you're setting the container up with has a `totalPage` of 1, so incrementing the page won't have any effect.

For this test, you should modify the `totalPage` count to be more than one and the current `page` to be 1. Then, call `nextTransactionPage` and verify that the `page` number changed.

You should also write a second test that verifies nothing happens when `nextTransactionPage` is called and we are currently on the last page.

Because `nextTransactionPage` calls `fetchTransactions` in turn, I would mock the entire `fetchTransactions` rather than trying to implement another promise/API mock and just verify that the right page number argument is being passed in.

tests/containers/award/AwardContainer-test.jsx
ejgullo: 👍

tests/containers/award/AwardContainer-test.jsx
ejgullo: 👍

PR #92, opened 2017-03-09
Award Amount Filter

mab6bf:
- User can select any or all of five `predefined` award amount ranges
- User can enter a `custom` award amount range
  - Clicking `custom` range button shows additional checkbox, with the user input (at the time the button was clicked) as the label
    - Clicking button again will overwrite the search and checkbox label with updated user input
    - Leaving `min` blank shows "Under $`max`"
    - Leaving `max` blank, or a `min` that is greater than a `max`, shows "$`min` & Above"
    - Sane inputs show "$`min` to $`max`"
- Selecting a `custom` award amount range will override any `predefined` award amount range selections, and vice-versa


src/js/helpers/awardAmountHelper.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+    else if (range[1] === 0) {
+        return `$${minLabel} & Above`;
+    }
+    return `$${minLabel} - $${maxLabel}`;
+};
+
+export const ensureInputIsNumeric = (input) => {
+    // Remove non-numeric characteres from input
+    const cleanInput = (input.toString()).replace(/\D/g, '');
```
kevinli-work: Unfortunately this also strips out decimal points, so `1.2` becomes `12`. AccountingJS has an [unformat()](http://openexchangerates.github.io/accounting.js/#documentation) function that may be helpful.

src/js/redux/reducers/search/filters/awardAmountFilterFunctions.js
```diff
@@ -0,0 +1,50 @@
...
+export const updateAwardAmounts = (state, value) => {
+    let updatedSet = state;
+
+    // value.amount will either be:
+    // 1) an integer, corresponding to included awardRanges
+    // 2) an array, with two items, corresponding to user input
+
+    if (value.searchType === 'specific') {
+        // Toggle specific range selection
+        if (updatedSet.get(`5`) !== undefined &&
```
mab6bf: Changing `5` to `specific` and the numbers to `range-[x]`

src/js/helpers/awardAmountHelper.js
mab6bf: Good call

PR #101, opened 2017-03-17
Grant Summary Page

kevinli-work:
* Refactored contract summary page so both pages are consistent
* Removed transaction table from AwardContainer, it is now its own container and the Award API promise is no longer combined
* Removed potential ceiling from grant visualization
* Centralized award type string logic
* Fixed bugs in the detail tables
* Added/refactored tests to handle new code architecture


src/js/components/award/financialAssistance/GrantDetails.jsx
```diff
@@ -0,0 +1,136 @@
...
+        }
+        if (award.description) {
+            description = award.description;
+        }
+        else {
+            description = "Not Available";
+        }
+
+        // CFDA Data
+        // TODO: get program description (objectives) for latest transaction
```
mab6bf: We need to hook this up later when the API has the data connected

PR #109, opened 2017-03-24
DS-888 Award Amounts Updates

ejgullo:
- Adding space to prevent cut-off amounts
- Adding link to contract details to slide down to the table (moving the clickTab function up to parent so that it can be updated by and passed down to the child components)
- Swapping our parent duns for parent id in summary bar
- Updating contract details to include the country name when only a state is present
- Updating alignment of current award amount labels
- Updating strings to proper terms for grants, and removing extraneous spaces


src/js/components/award/contract/ContractDetails.jsx
```diff
@@ -100,12 +102,19 @@ export default class ContractDetails extends React.Component {
             pricing = award.type_of_contract_pricing_description;
         }
 
+        // char count
+        let seeMore = false;
+        if (award.description.length > 160) {
```
mab6bf: Let's move this magical `160` number into class property and pass it down to `DetailRow`

src/_scss/pages/award/_moreButton.scss
```diff
@@ -0,0 +1,31 @@
...
+
+    &:visited {
+        color: $color-visited;
+    }
+
+    &:focus {
+        box-shadow: $focus-shadow;
+        outline: 0;
+    }
+    margin-top: .5rem;
```
mab6bf: Removing `margin-top: .5rem;` will keep the `line-height` consistent on the line in which the More/Less link is displayed

PR #110, opened 2017-03-27
Budget Category Filter - Sidebar and Top Bar Implementations

mab6bf:
- User can search and select Budget Functions and Federal Accounts through their respective Autocompletes
- User can select any combination of top-level object classes checkboxes
- Top Bar implementation of Budget Function, Federal Account, and Object Class selections
- Unit tests for the above
- A new SearchAccountOperation, which will eventually allow users to filter based on Budget Category. FY filters are the only other filters that apply to searches when Budget Category filters are used.

@kaitlin Since we weren't able to get the visualization changes done, do we want to save merging this PR until next sprint (as using these filters won't actually affect search results yet)?


kevinli-work:
There is a CSS bug with the error messages for budget function and federal account:
screen shot 2017-03-28 at 5 01 14 pm

PR #111, opened 2017-03-27
Federal Account Summary page

kevinli-work:
* Added Include Paths for Sass to allow for reasonable absolute (relative to `/_scss`) @include paths instead of large relative paths
* Implemented Redux actions and reducers to support federal account page
* Implemented federal account page scaffolding, header bar, and fiscal year text summary
* Implemented spending over time visualization with API and filter integration for federal account page (years only)
* Implemented spending by category visualization with API and filter integration for federal account page (program activity only)
* Implemented time filter and top filter bar
* Refactored award search spending over time visualization containers and components to allow for component reuse on federal account page
* Refactored award search spending by category visualization containers and components to allow for component reuse on federal account page
* Refactored top filter container and all child filter group components on award search page to handle filter removal at the filter group level rather than in the container
* Refactored the way that the filter sidebar is populated on award search page so the sidebar component could be reused on federal account page
* Refactored certain parts of the award search page Sass to allow for code reuse on federal account page


tests/containers/account/visualizations/AccountRankVisualizationContainer-test.jsx
```diff
@@ -0,0 +1,227 @@
...
+                reduxFilters={defaultFilters}
+                account={mockReduxAccount} />);
+
+            container.setState({
+                page: 5
+            });
+
+            expect(container.state().page).toEqual(5);
+
+            container.instance().changeScope('something else');
```
mab6bf: Do we have a real-ish example for this?

src/js/models/account/queries/AccountSearchBalanceOperation.js
```diff
@@ -0,0 +1,66 @@
+/**
+ * AccountSearchOperation.js
```
mab6bf: Is there a reason these names don't match?

tests/containers/account/visualizations/AccountRankVisualizationContainer-test.jsx
kevinli-work: No, because at time of development we don't know what the API values for other scopes would be (also the actual scope value is irrelevant to the logic of the test).

src/js/components/account/visualizations/time/TimeVisualization.jsx
```diff
@@ -0,0 +1,101 @@
...
+            tooltipData: null,
+            tooltipX: 0,
+            tooltipY: 0
+        };
+
+        this.showTooltip = this.showTooltip.bind(this);
+    }
+
+    showTooltip() {
+        // this.setState({
```
kevinli-work: It is placeholding for future code

src/js/components/account/visualizations/time/AccountTimeVisualizationSection.jsx
```diff
@@ -0,0 +1,108 @@
...
+            // width changed, update the visualization width
+            this.setState({
+                windowWidth,
+                visualizationWidth: this.sectionHr.offsetWidth
+            });
+        }
+    }
+
+    changePeriod(period) {
+        console.log(period);
```
kevinli-work: An empty function will cause the linter to fail

PR #115, opened 2017-03-29
Updates to the top buttons on the Homepage 

dfrasier:
* Fixed issue with the two buttons in the top section of the homepage.
* Added “Title” text to the two buttons in the top section of the homepage.
* Increased the padding of the button within the bottom section of the homepage.
* Added a hover transition tot the button within the bottom section of the homepage.


src/js/components/homepage/LinksSectionBox.jsx
```diff
@@ -41,16 +41,32 @@ export default class LinksSectionBox extends React.Component {
     }
 
     render() {
+        const arrow = (<Icons.AngleRight />);
```
mab6bf: No real need for setting this as a variable - we can move `<Icons.AngleRight />` into the <a> below

AGuyNamedMarco: Yup -- I removed this.  Thanks for the catch!

PR #127, opened 2017-04-03
Feature new pagination

ebdabbs:
Updated frontend pagination to use new pagination style for the rank visualization and infinite scrolling. Using has_next_page and has_previous_page booleans rather than total number of pages to handle and render previous/ next buttons. Updated test cases.


afrasier:
Let #datastore on slack know when this goes in, so we can pull in the backend changes at the same time!

src/js/components/search/visualizations/rank/RankVisualizationSection.jsx
```diff
@@ -73,15 +73,15 @@ export default class RankVisualizationSection extends React.Component {
         let disablePrev = false;
         let hidePager = '';
 
-        if (this.props.total < this.props.page + 1) {
+        if (!this.props.hasNextPage) {
```
mab6bf: Since this check is now on the truthiness of a boolean, we can just set `disableNext = !this.props.hasNextPage;` and `disablePrev = !this.props.hasPreviousPage;` above

src/js/components/search/visualizations/rank/RankVisualizationSection.jsx
```diff
@@ -73,15 +73,15 @@ export default class RankVisualizationSection extends React.Component {
...
             disableNext = true;
         }
 
-        if (this.props.page <= 1) {
+        if (!this.props.hasPreviousPage) {
             disablePrev = true;
         }
 
-        if (this.props.total < 1 || this.props.loading) {
+        if ((!this.props.hasPreviousPage && !this.props.hasNextPage) || this.props.loading) {
```
mab6bf: For readability, if we make the above changes to `disableNext` and `disablePrev`, this statement can now become `if ((disableNext && disablePrev) || this.props.loading)`

PR #128, opened 2017-04-04
DS-997 Treemap hover

ejgullo:
Added treemap hover functionality:
- Budget function descriptions added to json
- Descriptions parsed and displayed when the matching tile is hovered over
- Description is removed on mouse out
- Style updated to show yellow border on top 3 tiles on initial render
- Removed yellow border after first hover


src/js/components/homepage/visualizations/TreeMapCell.jsx
```diff
@@ -82,6 +89,26 @@ export default class TreeMapCell extends React.Component {
         });
     }
 
+    mouseIn(label, value) {
+        this.props.toggleTooltip(label, value);
+        this.setState({
```
kevinli-work: What is this `overlay` state being used for (see also constructor)?

src/js/components/homepage/visualizations/TreeMap.jsx
```diff
@@ -84,29 +96,90 @@ export default class TreeMap extends React.Component {
...
+            individualValue: value,
+            showOverlay: false
+        });
+        const newNodes = [];
+
+        // create a copy of finalNodes
+        // update the copy
+        this.state.finalNodes.forEach((i) => {
+            const originalNode = i;
+            const updatedNode = update(originalNode, {
```
kevinli-work: I'll be honest - I don't like this. You should never be directly manipulating props and this comes too close to doing that (though I recognize you're actually cloning the component and passing in new props during the clone process). Furthermore, I see no indication that `update` is safe for use with components...I'm pretty sure React components are not meant to be cloned/copied.

The easiest alternative is slightly more computationally expensive, but you would simply add a new if statement to the `componentWillReceiveProps` function to re-calculate the treemap when the overlay/category/etc states change and pass the new props down to the child treemap cells in the `buildTree` function.

A more complex approach would be to hold references to the child cell instances in your state and change their highlight props as necessary individually.

src/js/components/homepage/visualizations/TreeMap.jsx
```diff
@@ -39,6 +48,9 @@ export default class TreeMap extends React.Component {
         if (nextProps.categories.children.length > 0) {
             this.buildTree(nextProps.categories, nextProps.colors);
         }
+        this.setState({
```
ejgullo: It should be wrapped, will update!

src/js/components/homepage/visualizations/TreeMap.jsx
ejgullo: In my original approach, I wanted to re-run the `buildTree` function, but was concerned that running the entire function again would be too heavy.  Additionally, as the overlay state changes, it changes in `state`, so I wasn't sure that `componentWillReceiveProps` was appropriate.  Would an update in `componentDidUpdate` be appropriate?

src/js/components/homepage/visualizations/TreeMapCell.jsx
ejgullo: Leftover from an earlier approach, no longer necessary.  Removing.

PR #135, opened 2017-04-10
Update federal account spending over time

ebdabbs:
Added view by quarters functionality using new endpoints; setup for Object Class filter.


src/js/containers/account/visualizations/AccountTimeVisualizationContainer.jsx
```diff
@@ -103,52 +148,92 @@ export class AccountTimeVisualizationSectionContainer extends React.Component {
...
+                const quarters = [];
+                const yData = [];
+                data.forEach((balance, index) => {
+                    balance.data.results.forEach((group, index2) => {
+                        if (index == 0) {
+                            quarters.push(group.item + " Q" + group.submission__reporting_fiscal_quarter);
+                            yData.push({});
+                            yData[index2].outlay = parseFloat(group.aggregate);
+                        }
+                        // is there a more efficient way to do this?
```
kevinli-work: Yes there is. `this.balanceRequests` is an array of objects. For each API call that is made:
```
{
    cancel: // function to cancel the API call
    promise: // the JS promise that contains the API call
    type: "outlay" // a string indicating the data type that lines up with the property you are setting
}
```

So you can simply do:
```
const type = this.balanceRequests[index2].type;
yData[index2][type] = parseFloat(group.aggregate);
```

kevinli-work: Also, come up with a better variable name than `index2`.

src/js/helpers/accountQuartersHelper.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * accountQuartersHelper.js
+ * Created by Lizzie Salita 4/04/17
+ */
+
+import Axios, { CancelToken } from 'axios';
+
+import kGlobalConstants from 'GlobalConstants';
+
+export const fetchFederalAccount = (id) => {
```
kevinli-work: Is this a duplicate from `accountHelper`?

PR #136, opened 2017-04-10
DS-979 Award Summary Additional Details

ejgullo:
Updates for Financial System Information and Additional Information tables on Contract & Grant pages
- Added default 'Not Available' state to all fields
- Replaced null values with available fields
- Added values to `FinancialSystemItem.js` handler
- Remapped field names that have changed
- Verified fields with Kristine


src/js/dataMapping/contracts/additionalDetails.js
```diff
@@ -43,12 +43,12 @@ export const parentFields = [
     {
         label: 'IDC Type',
         field: '__special',
-        parse: () => null
+        parse: (data) => data.latest_transaction.contract_data.type_of_idc
```
kevinli-work: Unless you need to do custom parsing, you can just replace the `field` value with the field name and remove the `parse` function. These are mostly here to return `null` values before we had these fields in the API.

src/js/components/award/details/additional/AssistanceAdditionalDetails.jsx
```diff
@@ -14,12 +17,95 @@ const propTypes = {
...
+                let finItem = {};
+
+                res.data.results.forEach((item) => {
+                    finItem = new FinancialSystemItem(item);
+                });
+
+
+                this.setState({
+                    hasLoaded: true,
+                    inFlight: false,
```
kevinli-work: Can we use the `inFlight` state to prevent the group values from display "Not available" while the API request is still pending? Right now it flickers to "Not available" until the API responds.

src/js/dataMapping/contracts/additionalDetails.js
ejgullo: Gotcha, I'll clear those up!

kevinli-work:
This is happening when there's no date range:
screen shot 2017-04-11 at 11 18 00 am

Also can you deal with the merge conflicts?

PR #137, opened 2017-04-10
Feature About Page

dfrasier:
About us Page!


kevinli-work:
This PR appears to break the CSS on the awards pages:
screen shot 2017-04-11 at 11 21 13 am

src/_scss/pages/about/aboutPage.scss
```diff
@@ -0,0 +1,26 @@
+@import "../../mixins/fullSectionWrap";
```
kevinli-work: Import statements need to happen inside the CSS class block, otherwise the styles being imported are unconstrained - that is automatically applied throughout the site. This is causing other pages to display incorrectly.

PR #138, opened 2017-04-10
Fix: Object Class in Financial System

kevinli-work:
* Update Financial System Details model to correctly parse updated API response


kevinli-work:
* Updated to include corrected table pagination and updated tests

PR #143, opened 2017-04-14
Federal Accounts: Spending by Award

kevinli-work:
* Implements spending by award table on the federal account page
* Integrates with FY filter and object class filter
* Tables are sortable and link to their award pages


src/js/components/account/awards/AccountAwardsSection.jsx
```diff
@@ -0,0 +1,79 @@
...
+        window.addEventListener('resize', this.setTableWidth);
+    }
+
+    componentWillUnmount() {
+        // stop watching for size changes
+        window.removeEventListener('resize', this.setTableWidth);
+    }
+
+    setTableWidth() {
+        const tableWidth = this.tableWidthController.clientWidth - 2;
```
mab6bf: Can we document why this is required?

PR #153, opened 2017-04-20
Update federal account spending over time

ebdabbs:
See changes updated in the spending over time graph when data is filtered by Object Class on the Federal Account page.


src/js/containers/account/visualizations/AccountTimeVisualizationContainer.jsx
```diff
@@ -79,13 +79,85 @@ export class AccountTimeVisualizationSectionContainer extends React.Component {
...
+        // const filters = searchOperation.toParams();
+        const balanceFilters = searchOperation.toParams();
+        let filters = balanceFilters;
 
         const requests = [];
         const promises = [];
 
-        if (this.state.hasFilteredObligated) {
-            // In progress
+        const hasFilteredObligated = (this.props.reduxFilters.objectClass.count() > 0)
```
kevinli-work: Can we just pass `hasFilteredObligated` down to the child components as a prop instead of calculating it in every component?

src/js/containers/account/visualizations/AccountTimeVisualizationContainer.jsx
```diff
@@ -153,8 +223,93 @@ export class AccountTimeVisualizationSectionContainer extends React.Component {
         const xSeries = [];
         const ySeries = [];
         const allY = [];
-        if (this.state.hasFilteredObligated) {
-            // In progress
+        const hasFilteredObligated = (this.props.reduxFilters.objectClass.count() > 0)
```
kevinli-work: See note above; I feel like we should be able to hold this in state for this component rather than redoing the `if` every time

src/js/containers/account/visualizations/AccountTimeVisualizationContainer.jsx
```diff
@@ -153,8 +223,93 @@ export class AccountTimeVisualizationSectionContainer extends React.Component {
...
+            || (this.props.reduxFilters.programActivity.count() > 0);
+
+        if (hasFilteredObligated) {
+            if (this.state.visualizationPeriod === 'quarter') {
```
kevinli-work: Is there a way to combine the `hasFilteredObligated` true and false logic into one? It seems that both are doing nearly identical parsing with the exception of `true` calculating the `obligatedOther` value. Would it make more sense to combine the two sets of logic and just use the if-statement to determine whether or not to calculate `obligatedOther`?

src/js/containers/account/visualizations/AccountTimeVisualizationContainer.jsx
ebdabbs: Made hasFilteredObligated a state variable & passed down to child components as a prop

src/js/containers/account/visualizations/AccountTimeVisualizationContainer.jsx
ebdabbs: Made hasFilteredObligated a state variable & passed down to child components as a prop

src/js/containers/account/visualizations/AccountTimeVisualizationContainer.jsx
ebdabbs: Simplified `fetchData()` and `parseData()` logic by checking for `visualizationPeriod` and `hasFilteredObligated` only where necessary and eliminating repetitive if-else blocks.

PR #154, opened 2017-04-20
Subaward tab on Awards Page

kevinli-work:
* Implements the subaward tab on the awards page
* Allows column sorting and pagination
* Integrates with API


kevinli-work:
Sample IDs:
165 is a grant with subawards
1120 is a contract with subawards

PR #170, opened 2017-04-28
Add reason for modification to contract transaction table

ebdabbs:


src/js/dataMapping/contracts/transactionTable.js
```diff
@@ -5,33 +5,38 @@
 
 const tableSearchFields = {
     columnWidths: {
-        modification: 230,
-        action_date: 200,
-        amount: 220,
-        description: 380
+        modification: 184,
```
kevinli-work: Is there a reason the column widths were reduced? Having the columns too narrow will cause the header cells to wrap (which don't work correctly) when alternative fallbacks fonts are loaded (government computers are configured to ignore custom fonts). Can we restore the wider column widths?

kevinli-work: You can test this by replacing the `$font-sans` value in `_variables.scss` to just `$helvetica`:
screen shot 2017-05-02 at 3 18 02 pm

ebdabbs: Those widths were based on looking at the mockup (with custom font) - reverted to original values to prevent wrapping.

PR #177, opened 2017-05-01
Spending by Funding Agency and Spending by Recipient Visualizations

mab6bf:
- User can select Funding Agency from the Spending By dropdown on the Search page
  - User is presented with Agencies option by default
  - User can select from Agencies and Sub-Agencies options
  - Offices option is designated as "Coming Soon"
- User can select Recipient from the Spending By dropdown on the Search page
  - User is presented with Subsidiary option by default
  - Parent Company and Recipient Type are designated as "Coming Soon"


src/js/containers/search/visualizations/rank/SpendingByRecipientVisualizationContainer.jsx
```diff
@@ -0,0 +1,276 @@
...
+        this.fetchTransactions('Recipient vis - award filters');
+    }
+
+    fetchComboRequest() {
+        // a combination of budget and award filters have been selected
+        this.fetchAccountAwards('Recipient vis - combination');
+    }
+
+    fetchTransactions(auditTrail = null) {
+        const idField = FilterFields.accountAwardsFields.recipientId;
```
kevinli-work: Why are you pulling from `accountAwardsFields` in the transactions function? Why group by `award__recipient__xxx` instead of `recipient__xxx`?

mab6bf: Oversight - correcting now

PR #227, opened 2017-05-09
update warning banner to be clear about old data

wslack:


kaitlin:
I've submitted your changes to the powers at be for clearance :)

PR #228, opened 2017-05-09
[DS-1141] Subfunction Treemap

ejgullo:
Function Treemap:
- Treemap is now full width.  Information has moved from a sidebar to tooltips.

Subfunction Treemap:
- On click of a tile in the budget functions treemap (homepage), a second treemap, displaying that function's subfunctions appears.  It functions the same as the functions treemap does, including color change on hover, and tooltips for information.
- The original treemap is minimized, and has navigational arrows.  These include previous, next, and back.  The tile colors will fade, and the selected function tile is opaque.
- Mobile View is not supported at this time, as it is still in the design stage.


src/js/components/homepage/visualizations/treemap/TreeMap.jsx
```diff
@@ -6,15 +6,19 @@
...
 const propTypes = {
     categories: React.PropTypes.object,
     descriptions: React.PropTypes.array,
-    colors: React.PropTypes.array
+    colors: React.PropTypes.array,
+    subfunctions: React.PropTypes.object,
+    alternateColors: React.PropTypes.array
 };
 
 export default class TreeMap extends React.Component {
```
kevinli-work: I think this component has become too overloaded that it has become difficult to keep the logic straight. It might be more maintainable if you split this into two components (a high-level treemap and the subfunction treemap), then wrap those in a component that's just responsible for switching between them and passing the correct data down.

src/js/components/homepage/visualizations/treemap/TreeMapCell.jsx
```diff
@@ -127,25 +181,31 @@ export default class TreeMapCell extends React.Component {
         return (
             <g
```
kevinli-work: I think there needs to be a cursor icon change to indicate when these cells are clickable.

src/js/components/homepage/visualizations/treemap/TreeMap.jsx
```diff
@@ -67,27 +78,41 @@ export default class TreeMap extends React.Component {
...
         .sort((a, b) => b.height - a.height || b.value - a.value);
 
         // set up a treemap object and pass in the root
         let tileStyle = d3.treemapBinary;
+        let mapHeight = 565;
+        let mapWidth = this.state.visualizationWidth;
         if (this.state.windowWidth < 768) {
             tileStyle = d3.treemapSlice;
         }
+        if (sub === true) {
```
kevinli-work: The collapsed tree view in subtree mode should probably be its own component rather than a parameter of the normal treemap. Keeping it all as one monolith decreases the understandability and maintainability of the code.

src/js/components/homepage/visualizations/treemap/TreeMapTooltip.jsx
```diff
@@ -0,0 +1,121 @@
...
+        this.positionTooltip();
+    }
+    componentDidUpdate() {
+        this.positionTooltip();
+    }
+
+    positionTooltip() {
+        // we need to wait for the tooltip to render before we can full position it due to its
+        // dynamic width
+        const tooltipWidth = this.div.offsetWidth;
```
kevinli-work: I think there's some sizing and positioning issues with the tooltip on smaller screen sizes:
screen shot 2017-05-12 at 11 53 54 am
screen shot 2017-05-12 at 11 59 41 am

ejgullo: This will be handled with a separate PR to come including the new mobile functionality requested by designer.

src/js/components/homepage/visualizations/treemap/TreeMap.jsx
ejgullo: Agree, it's become a little scary.  Appreciate the suggestion on where to start separating it!

PR #235, opened 2017-05-11
[DS-1147] multiple recipients

ejgullo:
Update to check for multiple recipients with case insensitivity.

IDs to verify:
78304 on dev ("Multiple Recipients")
946692 on staging ("MULTIPLE RECIPIENTS")


kevinli-work:
Moved to #237

PR #250, opened 2017-05-19
Add/ Remove Award Table Columns

ebdabbs:
- Added button and dropdown for selecting which columns to display
- Store column visibility in the Redux state
    - Redux action and reducer for toggling a column's visibility
- Added 63 new columns to the contracts tab
    - Script to generate `tableSearchFields` data mapping and `AwardSummary` model code and corresponding csv file are available in JIRA subtask [DS-1175].
- Updated tests


src/js/redux/reducers/search/columnVisibilityReducer.js
```diff
@@ -0,0 +1,251 @@
+/**
+ * columnVisibilityReducer.js
+ * Created by Lizzie Salita 5/04/17
+ **/
+
+import { Record, OrderedSet } from 'immutable';
+
+import * as ColumnVisibilityFuncs from './filters/ColumnVisibilityFunctions';
+
+const initialState = {
```
kevinli-work: If we're going to have 60+ columns, let's move this into its own file. Also if you `export` the initialState, you can use it in your tests without having to copy it over there and maintain two versions

src/js/containers/search/table/ResultsTableContainer.jsx
```diff
@@ -116,6 +120,10 @@ export class ResultsTableContainer extends React.Component {
                 this.performSearch();
             }
         }
+        else if (prevProps.columnVisibility !== this.props.columnVisibility) {
+            // Visible columns have changed
+            this.showColumns(this.props.meta.tableType);
```
kevinli-work: It seems that the container only updates the table UI when it receives new columns from Redux, however it doesn't trigger a new API call. Right now, if you look in the `performSearch` function, the API only receives a limited set of fields (specified in `_requestFields` in the dataMapping file). As you add/remove columns, new searches should be performed to add these fields into the API request. You'll also want to "reset" the search by passing in `true` to the newSearch param in performSearch so that it will reset the table (returning it to the top left), that way if the user is on page 2+, they won't be missing the data for already loaded pages.

src/js/components/search/table/ResultsTable.jsx
```diff
@@ -41,9 +41,11 @@ export default class ResultsTable extends React.PureComponent {
     }
 
     componentWillReceiveProps(nextProps) {
+        const currentType = nextProps.currentType;
+        const visibileCols = parseFloat(nextProps.columns.length);
```
kevinli-work: 1. `visible` is spelled incorrectly.
2. Is getting a count of the column length sufficient to generate a hash with? It may be the case that the user removes an existing column and adds a new one - new data but the count is different. Many ImmutableJS objects can give you [hashCodes](https://facebook.github.io/immutable-js/docs/#/OrderedSet/hashCode) to check for equality that may be useful for this.

kevinli-work:
Also, what is the behavior of the table sorting order when the active sort column is removed?

src/js/redux/reducers/search/columnVisibilityReducer.js
ebdabbs: Moved to `src/js/redux/reducers/search/columnVisibilityReducerInitialState.js`

src/js/containers/search/table/ResultsTableContainer.jsx
ebdabbs: Now performs a new search when columns are added or removed, requesting only the fields pertaining to visible columns.

src/js/components/search/table/ResultsTable.jsx
ebdabbs: Fixed typo; replaced length with hashCode() on OrderedSet of visible columns

ebdabbs:
@kevinli-work currently it stays in the same order until the user selects another sort order. Spoke with Geoff and he recommended leaving it that way since it would be awkward to resort when a column is removed.

PR #258, opened 2017-05-30
[DS-1296] Shared Checkbox Used Across Site

ejgullo:
- Utilizing new shared checkbox component in:
1. Award Type filter on Search Page
2. Object Class filter on Account Page
3. Program Activity filter on Account Page

- All Checkbox filters don't automatically expand to show sub-tiers when a top-tier box is checked


src/js/components/account/filters/objectClass/ObjectClassFilter.jsx
```diff
@@ -22,28 +22,33 @@ export default class ObjectClassFilter extends React.Component {
...
-                code={code}
-                label={label}
-                checked={checked}
-                toggleValue={this.toggleValue} />);
+            return (<PrimaryCheckboxType
+                {...code}
+                {...this.props}
+                name={label}
+                value={code}
+                key={index}
```
kevinli-work: If `code` is unique (which I think it is, since the original implementation used it as the `key`), then it is [better to use](https://facebook.github.io/react/docs/reconciliation.html#keys) that as the `key` value rather than an index number. This is because the React can associate a specific data element to a component and watch both of those for changes. When an iterated number is used, React is more likely to re-render the component.

src/_scss/pages/search/filters/awardType/awardType.scss
```diff
@@ -3,11 +3,52 @@
...
     }
+    @import '../location/_selectedButtons';
 
-    @import "./_primaryType";
-
-    ul.secondary-award-set {
-        @include unstyled-list;
-        @import "./_secondaryType";
+    .recipient-radio,
+    .checkbox-type-filter {
```
kevinli-work: Please review the padding on this filter compared to the other filters:
screen shot 2017-06-01 at 1 33 28 pm

src/js/components/account/filters/programActivity/ProgramActivityFilter.jsx
```diff
@@ -65,17 +65,22 @@ export default class ProgramActivityFilter extends React.Component {
...
+                            {...programActivity}
+                            {...this.props}
+                            name={label}
+                            value={programActivity.id}
                             key={programActivity.id}
-                            programActivityID={programActivity.id}
-                            label={label}
-                            checked={checked}
-                            toggleValue={this.toggleValue} />);
+                            types={this.props.availableProgramActivities}
```
kevinli-work: `types` is supposed to be an object - see the JS console error message that comes up whenever the federal account page is loaded

src/js/components/sharedComponents/checkbox/PrimaryCheckboxType.jsx
```diff
@@ -111,18 +111,16 @@ export default class PrimaryCheckboxType extends React.Component {
     }
 
     toggleChildren() {
-        let showChildren = true;
-        let arrowState = 'expanded';
+        // TO-DO: Fix input for bulkTypeChange on award type filter
```
kevinli-work: 1. todos require your name on them so we know who wrote it in the future
2. I don't understand this bug, please explain it further

kevinli-work:
Also: please fix the parent checkbox behavior for items with children. When a parent is selected, it should automatically expand to show the selected children. Currently it doesn't do that, but it _does_ disable the expand button, so it is impossible to view the children without first deselecting.

screen shot 2017-06-01 at 1 45 56 pm

ejgullo:
@kevinli-work To your last comment, that change is to address this point of the A/C for the story:

- All Checkbox filters don't automatically expand to show sub-tiers when a top-tier box is checked

kevinli-work:
@ejgullo @dfrasier Okay, but I feel like the expand button shouldn't be disabled then.

src/js/components/sharedComponents/checkbox/PrimaryCheckboxType.jsx
ejgullo: Irrelevant now, removing.

src/js/components/account/filters/programActivity/ProgramActivityFilter.jsx
ejgullo: would converting to an object here via `_.keyBy` be acceptable?

src/js/components/account/filters/objectClass/ObjectClassFilter.jsx
ejgullo: 👍

PR #284, opened 2017-06-13
[DS-927] Adding subobject classes to search filter

ejgullo:
- Added second tier of object classes to search filter, using standardized checkbox components
- Updated search field to properly use the subobject codes
- Updated tests to allow for the new use of Set rather than ObjectMap


src/js/components/search/filters/budgetCategory/BudgetCategoryOCSearch.jsx
```diff
@@ -4,34 +4,61 @@
...
-        Object.keys(objectClassDefinitions).forEach((key) => {
-            objectClassItems.push(
+        const objectClassItems = (
+            this.props.objectClassMapping.map((type, index) =>
                 <PrimaryCheckboxType
+                    {...type}
                     {...this.props}
-                    key={key}
-                    id={`award-${key}`}
+                    key={index}
```
kevinli-work: The `objectClassMapping` already has unique values that can be better utilized as `key`. Again, using an iterator index as `key` should be a last option.

ejgullo: 👍

src/js/containers/search/filters/budgetCategory/BudgetCategoryOCContainer.jsx
```diff
@@ -13,15 +13,17 @@ import BudgetCategoryOCSearch from
...
+    bulkObjectClassesChange: React.PropTypes.func
 };
 
 export class BudgetCategoryOCContainer extends React.Component {
     render() {
         return (
             <BudgetCategoryOCSearch
                 {...this.props}
-                selectObjectClass={this.props.updateObjectClasses.bind(this)} />
+                selectObjectClass={this.props.updateObjectClasses.bind(this)}
```
ejgullo: 👍

geoffgeiling:
Design review complete and approved.
*One small issue I discovered in design review and communicated to Emily within slack was that the object sub-class "full time permanent" was showing when chosen as a filter. We discovered it was a bad mapping and the issue is fixed now

PR #294, opened 2017-06-15
Fix Map Scrolling on Mobile

mab6bf:
- Map cannot be panned on viewports smaller than 768px, allowing user to scroll down the entire page
- In viewports smaller than 768px, buttons appear at the bottom right of the map that allow user to navigate up, down, left, and right 100px at a time

### DIFF
--- a/src/js/components/search/visualizations/geo/map/MapBox.jsx
+++ b/src/js/components/search/visualizations/geo/map/MapBox.jsx
@@ -5,7 +5,7 @@
 
 import React from 'react';
 import MapboxGL from 'mapbox-gl/dist/mapbox-gl';
-import _ from 'lodash';
+import { throttle } from 'lodash';
 import * as Icons from 'components/sharedComponents/icons/Icons';
 
 import kGlobalConstants from 'GlobalConstants';
@@ -17,6 +17,7 @@ const propTypes = {
     hideTooltip: React.PropTypes.func
 };
 
+// Define map movement increment
 const delta = 100;
 
 export default class MapBox extends React.Component {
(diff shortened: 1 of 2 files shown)


src/js/components/search/visualizations/geo/map/MapBox.jsx
```diff
@@ -5,6 +5,8 @@
 
 import React from 'react';
 import MapboxGL from 'mapbox-gl/dist/mapbox-gl';
+import _ from 'lodash';
```
kevinli-work: Can you import just `{ throttle }`?

src/js/components/search/visualizations/geo/map/MapBox.jsx
```diff
@@ -134,13 +215,67 @@ export default class MapBox extends React.Component {
...
         return (
             <div
                 className="mapbox-item"
                 ref={(div) => {
                     this.mapDiv = div;
-                }} />
+                }}>
+                <div className={`map-buttons ${hideClass}`}>
+                    <div className="first-row">
+                        <button
```
kevinli-work: Is there a reason for accessing the buttons via refs instead of setting up `onMouseDown`/`onMouseUp`/`onTouchStart`/`onTouchEnd` props?

src/js/components/search/visualizations/geo/map/MapBox.jsx
```diff
@@ -56,6 +68,67 @@ export default class MapBox extends React.Component {
...
+            case 'down':
+                map.panBy([0, delta]);
+                break;
+            default:
+                break;
+        }
+    }
+
+    buttonListen(button, map) {
+        this.persist = setInterval(() => {
```
kevinli-work: Why are the button events timers? Is it so you can hold them down and constantly pan?
Also, if they must be timers, can you ensure they check for existing timers and cancel them? If a user brings up 3D Touch or a context menu, the map begins to infinitely scroll.

src/js/components/search/visualizations/geo/map/MapBox.jsx
mab6bf: Yep

src/js/components/search/visualizations/geo/map/MapBox.jsx
mab6bf: Removed timers

src/js/components/search/visualizations/geo/map/MapBox.jsx
mab6bf: Fixed

PR #296, opened 2017-06-15
Agency Page

mab6bf:
- User can see base-level Agency info that is currently available from the API
- User can see the Obligated Amount visualization
- User can see the Object Classes visualization
- User can see the Recipients visualization ("All" option)

NB: Please test with `localhost:3000/#/agency/256`


geoffgeiling:
Design review complete. Changes discussed via slack and made include:
1. realigning text on top of object class treemap
2. changes to the object class tooltips
3. transition wording corrections
4. Kevin created a story to allow for sorting in the recipient visualization

In the future, will use comments in GitHub to discuss changes

PR #297, opened 2017-06-16
[DS-1338] Address Display - Award Page Recipient & PoP

ejgullo:
- Reformatting the address display in recipient & pop fields for individual award pages
- Met the A/C, PM is aware that some values may look odd because we are missing some data, acceptable for the time being.


src/js/components/award/financialAssistance/FinancialAssistanceDetails.jsx
```diff
@@ -61,24 +61,13 @@ export default class FinancialAssistanceDetails extends React.Component {
 
         // Location
         let popPlace = "Not Available";
-        let popZip = null;
-        if (award.pop_zip) {
-            popZip = award.pop_zip;
+        if (award.pop_country_code === 'USA') {
+            popPlace = `${award.pop_city}, ${award.pop_state_province} ${award.pop_zip}`;
```
kevinli-work: Why is city but no state/state but no city not being handled in this view?

src/js/components/award/contract/ContractDetails.jsx
```diff
@@ -74,24 +74,25 @@ export default class ContractDetails extends React.Component {
...
-            popPlace = `${award.pop_city} ${popZip}`;
+        else if (stateProvince) {
+            cityState = stateProvince;
         }
-        else if (award.pop_city && !award.pop_state_province && !popZip) {
-            popPlace = award.pop_city;
+        if (award.pop_country_code === 'USA') {
+            popPlace = `${cityState} ${award.pop_zip}`;
+            popPlace +=
+            `\nCongressional District: ${award.pop_state_code}-${award.pop_congressional_district}`;
```
kevinli-work: What's the expected behavior when state code and/or congressional district isn't available?

src/js/components/award/financialAssistance/FinancialAssistanceDetails.jsx
```diff
@@ -61,24 +61,13 @@ export default class FinancialAssistanceDetails extends React.Component {
...
+            popPlace +=
+            `\nCongressional District: ${award.pop_state_code}-${award.pop_congressional_district}`;
```
kevinli-work: Same, what if there is no state code or congressional district (see `award/603`)?

src/js/components/award/financialAssistance/FinancialAssistanceDetails.jsx
ejgullo: forgot to update this file with this morning's new requests - fixing now

src/js/components/award/financialAssistance/FinancialAssistanceDetails.jsx
ejgullo: same as above

src/js/components/award/contract/ContractDetails.jsx
ejgullo: Confirmed with kaitlin, we will be hiding it.  I'll add this logic in.

geoffgeiling:
Design review conducted and PR approved

PR #302, opened 2017-06-19
Autosized Table Columns

kevinli-work:
* Table columns can now be autosized based on their display label
* Optionally, when a wider table column is desired, it can be overridden programmatically
* When fallback fonts are used, text width calculations use those fonts
* Ported over Search Page awards table, Financial Accounts awards table, Contract/Financial Assistance/Loan transaction table, subawards table, and financial system details table
* Home page state table still uses static widths


src/js/helpers/textMeasurement.js
```diff
@@ -0,0 +1,40 @@
...
+
+    // iterate through the provided columns and measure their titles
+    const context = canvas.getContext('2d');
+    context.font = font;
+    return Math.ceil(context.measureText(text).width);
+};
+
+export const measureTableHeader = (text) => {
+    const tableHeaderFont = 'bold 14px Source Sans Pro, sans serif';
+    return measureText(tableHeaderFont, text) + 120;
```
mab6bf: What is 120? Can we document it?

PR #307, opened 2017-06-22
Responsive Tabs

ebdabbs:
- In mobile, tabs for switching table type become a dropdown list
- Applied to tables on Search, Federal Account, and Award Summary pages
- Refactoring to improve code reuse


src/js/components/award/details/DetailsTabBar.jsx
```diff
@@ -10,46 +10,20 @@ import DetailsTabItem from './DetailsTabItem';
...
-                label: 'Additional Details',
-                code: 'additional',
-                disabled: false
-            });
-        }
-        const tabs = tabsSet.map((tab) => (
+        const tabs = this.props.tabOptions.map((tab) => (
             <DetailsTabItem
                 {...tab}
                 active={tab.code === this.props.activeTab}
```
kevinli-work: The tabs are no longer highlighting because this needs to be changed over to `internal`

src/js/components/award/details/DetailsSection.jsx
```diff
@@ -97,12 +98,37 @@ export default class DetailsSection extends React.Component {
     render() {
         const content = this.currentSection();
 
+        const tabs = [
```
kevinli-work: Is there a reason this is in the `render` function instead of outside the class?

src/js/components/award/details/DetailsTabBar.jsx
ebdabbs: Changed `code` to `internal`

src/js/components/award/details/DetailsSection.jsx
ebdabbs: Moved tabs array outside of the class

PR #314, opened 2017-06-28
Update analytics

ebdabbs:
- Added Google Analytics to more search page filters
    - Budget Function
    - Federal Account
    - Award Amount (range)
    - Award ID (piid, uri, fain)
- Fixed Keyword filter analytics to show up in correct category
- Added analytics to the Glossary
    - Compare how many users open the glossary through the navigation link vs floating button
    - Log which glossary links are clicked on


src/js/containers/search/filters/KeywordContainer.jsx
```diff
@@ -20,18 +20,10 @@ const ga = require('react-ga');
...
-        });
-    }
-
     static logSelectedKeywordEvent(keyword) {
         ga.event({
-            category: 'Search Filters',
-            action: 'Selected Keyword',
+            category: 'Search Page Filter Applied',
+            action: 'Applied Keyword Filter',
             label: keyword
```
mab6bf: If I clear the box and press Submit, we'll be logging an empty string. Should we check if the keyword isn't null?

src/js/containers/search/filters/budgetCategory/BudgetCategorySearchContainer.jsx
```diff
@@ -18,7 +18,26 @@ const propTypes = {
...
 };
 
+
+const ga = require('react-ga');
+
 export class BudgetCategorySearchContainer extends React.Component {
+    static logBudgetFunctionFilterEvent(type) {
+        ga.event({
+            category: 'Search Page Filter Applied',
+            action: `Applied Budget Function Filter`,
```
mab6bf: We can use single quotes here, if this is a static string

src/js/containers/search/filters/budgetCategory/BudgetCategorySearchContainer.jsx
```diff
@@ -18,7 +18,26 @@ const propTypes = {
...
+            label: type
+        });
+    }
+
+    static logFederalAccountFilterEvent(type) {
+        ga.event({
+            category: 'Search Page Filter Applied',
+            action: `Applied Federal Account Filter`,
```
mab6bf: We can use single quotes here, if this is a static string

src/js/containers/search/filters/budgetCategory/BudgetCategorySearchContainer.jsx
```diff
@@ -30,10 +49,16 @@ export class BudgetCategorySearchContainer extends React.Component {
...
+
+        // Analytics
+        BudgetCategorySearchContainer.logBudgetFunctionFilterEvent(budgetFunction.title);
     }
 
     updateFederalAccounts(federalAccount) {
         this.props.updateSelectedFederalAccounts(federalAccount);
+
+        // Analytics
+        BudgetCategorySearchContainer.logFederalAccountFilterEvent(federalAccount.federal_account_code);
```
mab6bf: Would we want to turn these into a single `logFilterEvent` that would take both the type of event as well as the budget function title / federal account code?

src/js/containers/search/filters/recipient/RecipientSearchContainer.jsx
```diff
@@ -74,10 +74,12 @@ export class RecipientSearchContainer extends React.Component {
 
     toggleRecipientType(selection) {
         this.props.toggleRecipientType(selection);
+        // Analytics handled by checkbox component
```
mab6bf: +1 for documenting this

src/js/containers/search/filters/KeywordContainer.jsx
ebdabbs: Added a check on line 71

src/js/containers/search/filters/budgetCategory/BudgetCategorySearchContainer.jsx
ebdabbs: Combined the two budget category events and changed this to a dynamic string with an `event` parameter for Budget Function/ Federal Account

src/js/containers/search/filters/budgetCategory/BudgetCategorySearchContainer.jsx
ebdabbs: Changed to single quotes

src/js/components/search/filters/awardAmount/SpecificAwardAmountItem.jsx
```diff
@@ -12,7 +12,17 @@ const propTypes = {
     searchSpecificRange: React.PropTypes.func
 };
 
+const ga = require('react-ga');
+
 export default class SpecificAwardAmountItem extends React.Component {
+    static logAmountRangeEvent(range) {
+        ga.event({
+            category: 'Search Page Filter Applied',
+            action: `Applied Award Amount Range Filter`,
```
kevinli-work: So the tick marks are an ES6 feature called [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) which are used to inject dynamic content (variables) inside an otherwise static string. In most cases where you're just using a string, you should use the single quotes (in this case, `action`).

src/js/containers/search/filters/budgetCategory/BudgetCategorySearchContainer.jsx
kevinli-work: Missing `action`

src/js/containers/search/filters/budgetCategory/BudgetCategorySearchContainer.jsx
kevinli-work: Missing `action`

PR #317, opened 2017-06-29
Fix: Fix federal account filter bar style

kevinli-work:
* Fixes the federal account filter sidebar
* Fixes longstanding "Filter by" alignment issue on IE


geoffgeiling:
This approved per Design Review. However, in looking at this PR, I noticed that the Object Class filter is decoupled from the hierarchy. This appears to also be the case on staging already. We might need a story to fix this next sprint.

PR #318, opened 2017-06-30
Include only visible columns in the CSV download

ebdabbs:
- API call requests nested fields rather than top-level objects
- Separated columns that required multiple fields
- Removed request fields from data mapping since they are generated from visible columns


geoffgeiling:
Approved from design perspective. However, while not part of a/c, would like to talk about column header labels in sprint review. Is currently showing a mix of what appears to be API labels and UI labels

screen shot 2017-07-03 at 10 06 14 am

geoffgeiling:
Actually back that up... when I try downloading with all the columns added, the file is not generated (been running for 30+ minutes). Lizzie and I are wondering if maybe it's getting hung up on a specific column. I tried adding a just a few columns and it worked.

I am generating the file with only one row so there should not be any issue with it being a large data set.

ebdabbs:
Resolved the issue where the file was not generated by removing the "Domestic or Foreign Entity" column, which was mapped to a nonexistent field.

PR #320, opened 2017-06-30
Award table count badges

kevinli-work:
* Shows the number of awards in each tab on the search page
* Also implemented on the federal account detail page
* Uses old `/v1/awards/total/` API endpoint


geoffgeiling:
Looks great! Design approved

PR #328, opened 2017-07-07
Fix - Use Budget Category Filters in Awards Table Results

mab6bf:
- User selections of Budget Function, Federal Account, and Object Class filters will be reflected in the Awards table


geoffgeiling:
Design review approved

src/js/models/search/SearchOperation.js
```diff
@@ -115,6 +116,22 @@ class SearchOperation {
                 this.locationDomesticForeign, this.searchContext));
         }
 
+        // Add Budget Category queries
```
kevinli-work: If we add these to `commonParams`, let's take them out of the `uniqueParams` in the subclassed Search Operations.

PR #330, opened 2017-07-11
Agency Page - Budget Authority

mab6bf:
Adds the following budget authority information:
- Submission Period End Date
- Current year Budget Authority
- Current year Percentage of US Budget
- Percentage of US Budget Visualization Bar

Wording changes may follow later this week, and will be submitted in a subsequent PR.


src/js/GlobalConstants_old.js
```diff
@@ -0,0 +1,13 @@
+// Development values
```
kevinli-work: Please remove

src/js/components/agency/overview/AgencyOverview.jsx
```diff
@@ -23,6 +30,44 @@ export default class AgencyOverview extends React.Component {
             logo = (<img src={this.props.agency.logo} alt={this.props.agency.name} />);
         }
 
+        // Move props to variables for readability
```
kevinli-work: Can we move a lot of this to `componentWillReceiveProps` and store the calculated values in state?

src/js/components/agency/overview/AgencyOverview.jsx
```diff
@@ -23,6 +30,44 @@ export default class AgencyOverview extends React.Component {
...
+        const percentage = MoneyFormatter.calculateTreemapPercentage(
+            budgetAuthority, federalBudget);
+        const percentageElement = (
+            <span className="authority-statement-percentage">{percentage}</span>
+        );
+
+        // Generate visualization parameters
+        let visualizationWidth = 0;
+        if (this.containerDiv) {
+            visualizationWidth = this.containerDiv.getBoundingClientRect().width;
```
kevinli-work: This measures the div of the last render cycle - this currently works because of a bug in the sticky sidebar that triggers a re-render on every scroll event, so the measurement occurs extremely frequently. The scroll bug is fixed in #329 which means this measurement occurs on mount (but does nothing because the div doesn't exist yet) and after the API call is parsed (which gives it a good initial reading), but won't update again on browser resize.

A better way is to measure after `componentDidMount` (which is triggered after the initial `render`), then attach a window resize event to measure on resize. Store the width in state and use that value.

geoffgeiling:
Design review approved

src/js/GlobalConstants_old.js
mab6bf: Done

src/js/components/agency/overview/AgencyOverview.jsx
mab6bf: Yes

src/js/components/agency/overview/AgencyOverview.jsx
mab6bf: Fixed

PR #342, opened 2017-07-20
Other Award Types

mab6bf:
- User can see "Other" is a top-level option in the Award Type filter, with "Insurance" and "Other Financial Assistance" as children (https://projects.invisionapp.com/share/XHB29AYB3#/screens/205404741_USA_Spending_Search)
- User can see "Other" as an option in the tabs above the award table on the Search/Download and Federal Account pages, replacing "Insurance"
- User can see "Other" as a future option in the Agency Page Rank Visualization, replacing "Insurance"


geoffgeiling:
Design approved

PR #362, opened 2017-08-07
Sidebar Remix

mab6bf:
- Sidebar filters are now collapsed by default on *both* desktop and mobile
- Removed Budget Categories sidebar filter, reclassifying Funding Agency and Fiscal Year as Award-type filters, leaving only Award-type filters in the Sidebar
- Reorders the Sidebar filters in the order of most-used:
  - Keywords
  - Time Period
  - Award Types
  - Agencies
  - Recipients
  - Place of Performance
  - Award Amount
  - Award ID
  - [Other Awards Items]
- Other Award Items will be broken out into individual top-level sidebar containers in a following PR from @ejgullo

### DIFF
--- a/src/js/components/search/SearchPage.jsx
+++ b/src/js/components/search/SearchPage.jsx
@@ -17,7 +17,8 @@ import SearchSidebar from './SearchSidebar';
 import SearchResults from './SearchResults';
 
 const propTypes = {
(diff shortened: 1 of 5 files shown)


geoffgeiling:
Design approved

mab6bf:
Confirmed - adding that in soon

PR #363, opened 2017-08-09
Update Contract Summary Business Types

ebdabbs:
- Show a list of applicable business types on the Contract Summary page, instead of `Unknown Types`
- If there are more than two types, hide the overflow by default and give users the option to see more


geoffgeiling:
Design review approved

PR #364, opened 2017-08-09
New SearchOperation for Award Search page

mab6bf:


src/js/models/search/SearchAwardsOperation.js
```diff
@@ -0,0 +1,241 @@
...
+        if (this.keyword !== '') {
+            filters[rootKeys.keyword] = this.keyword;
+        }
+
+        // Add Time Period
+        if (this.timePeriodFY.length > 0 || this.timePeriodRange.length === 2) {
+            if (this.timePeriodType === 'fy' && this.timePeriodFY.length > 0) {
+                filters[rootKeys.timePeriod] = this.timePeriodFY.map((fy) =>
+                    ({
+                        [timePeriodKeys.startDate]: `${fy - 1}-10-01`,
```
kevinli-work: Can you generate these using the `fiscalYearHelper`?

src/js/models/search/SearchAwardsOperation.js
```diff
@@ -0,0 +1,241 @@
...
+                agencies.push({
+                    [agencyKeys.type]: 'funding',
+                    [agencyKeys.tier]: 'toptier',
+                    [agencyKeys.name]: agencyArray.toptier_agency.name
+                });
+            });
+
+            // Awarding Agencies can be both toptier and subtier
+            this.awardingAgencies.forEach((agencyArray) => {
+                const agencyName = agencyArray.agencyType === 'toptier'
```
kevinli-work: Should we do `agencyName = agencyArray[`${agencyArray.agencyType}_agency`].name` so it can handle offices in the future?

src/js/models/search/SearchAwardsOperation.js
```diff
@@ -0,0 +1,241 @@
...
+
+        this.selectedRecipients = state.selectedRecipients.toArray();
+        this.recipientDomesticForeign = state.recipientDomesticForeign;
+        this.selectedRecipientLocations = state.selectedRecipientLocations.toArray();
+        this.recipientType = state.recipientType.toArray();
+
+        this.selectedLocations = state.selectedLocations.toArray();
+        this.locationDomesticForeign = state.locationDomesticForeign;
+
+        this.awardAmounts = state.awardAmounts.toArray();
```
kevinli-work: This does not appear to account for custom ranges, which is a nested object in `awardAmounts`

src/js/models/search/SearchAwardsOperation.js
```diff
@@ -0,0 +1,241 @@
...
+            const amounts = [];
+
+            this.awardAmounts.forEach((awardAmount) => {
+                const amount = {};
+
+                if (awardAmount[0] !== 0) {
+                    amount[awardAmountKeys.min] = awardAmount[0];
+                }
+
+                if (awardAmount[1] !== 0) {
```
kevinli-work: Checking against 0 is going to do weird things with negatives, especially for custom ranges

src/js/models/search/SearchAwardsOperation.js
mab6bf: Yep, done

src/js/models/search/SearchAwardsOperation.js
mab6bf: Absolutely - done

src/js/models/search/SearchAwardsOperation.js
mab6bf: The backend is asking **only** for defined elements, which is why this differs a bit from the previous implementation. I can add the check to delete the max number if the min number is greater than it, similar to before.

src/js/models/search/SearchAwardsOperation.js
mab6bf: It does account for custom ranges

PR #366, opened 2017-08-10
Add federal account column to award detail pages

kevinli-work:
* Financial System Details table on individual award pages now has a federal account column
* Federal account column shows the name of the column and links to the federal account page


geoffgeiling:
design approved

PR #369, opened 2017-08-11
Fix white/gray row shading

ebdabbs:
- Implements alternating white & gray rows on the agency landing page table


geoffgeiling:
design review approved

PR #372, opened 2017-08-15
Fix agency obligated date

ebdabbs:
- "As of" date under the Obligated Amount visualization should match the date in the Overview section
- Obligated values are displayed with 1 decimal point precision


geoffgeiling:
Design review - approved

PR #375, opened 2017-08-17
Agency Page design updates

ebdabbs:
- "Data as of X" is added to all sections according to the mock (Date is the same for all visualizations except recipients, which shows the date awards were last uploaded)
- Labels added to each visualization so the users know they are looking at obligated amounts
- Asterisk is removed from obligations section and previous "data as of" note is removed


geoffgeiling:
Design Review - approved

PR #393, opened 2017-09-05
Add Agency Outlay

ebdabbs:
- Added outlay line to Agency Profile Page obligated amount visualization
- Updated the legend to show "Obligated Amount" instead of "Obligations"


geoffgeiling:
Design Review - Approved

src/js/components/agency/visualizations/obligated/ObligatedGraph.jsx
```diff
@@ -22,14 +24,46 @@ export default class AgencyObligatedGraph extends React.Component {
...
         const remainderValue = authorityValue - obligatedValue;
+        const outlayValue = this.props.outlay;
 
         let obligatedPercent = 0;
+        let outlayPercent = 0;
+        let remainderPercent = 0;
 
-        if (obligatedValue !== 0 && authorityValue !== 0) {
-            obligatedPercent = (obligatedValue / authorityValue);
+        if ((obligatedValue > 0) && (outlayValue > 0) && (authorityValue > 0)) {
```
kevinli-work: This seems like a pretty confusing way to approach this problem, given the large number of conditionals.

It looks like what we are essentially doing for authority, obligated, and outlay is dividing their value by the max value of the three or returning 0 if the value is negative. Why not just do that? `Math.max` to get the max value, divide each by the max, then use `Math.min` to ensure that negative values are rounded up to 0. The largest value, being divided by itself, will still come out to 1.

src/js/containers/agency/visualizations/ObligatedContainer.jsx
```diff
@@ -64,7 +65,8 @@ export class ObligatedContainer extends React.PureComponent {
 
                 this.setState({
```
kevinli-work: It worries me that this doesn't break our tests. This indicates this is either not tested (so we should add tests) or it's not testing the right thing (ie we should fix it)...

src/js/components/agency/visualizations/obligated/ObligatedGraph.jsx
ebdabbs: Simplified this logic using `Math.max`

src/js/containers/agency/visualizations/ObligatedContainer.jsx
ebdabbs: Added a new test that checks that the state variables are set correctly, including outlay amount

PR #394, opened 2017-09-06
Consistent obligated amount values

ebdabbs:
- Agency profile page shows consistent obligated amounts from section to section, that include negative obligations
- Obligated amounts are formatted with one decimal point precision


kevinli-work:
It looks like we're still passing the positives-only total obligation value down to the treemaps themselves, which are used for the tooltip percentage calculation in major object class and used for the tooltip percentage + the header total value in minor object class:
screen shot 2017-09-07 at 11 48 05 am

Do we want to use the positives-only total obligation or the actual obligation (the new `displayTotal` value) in these cases?

ebdabbs:
Updated the Object Class treemap to use actual obligation (including negative values) and added a note when the cells add up to over 100%.

geoffgeiling:
Design Review - Approved

PR #396, opened 2017-09-11
Dynamic Federal Account Object Class filter options

kevinli-work:
* Object class filter in federal account page only shows those that apply to the current federal account
* Object class filters are organized in major/minor hierarchy
* Federal account top filter bar now correctly counts the number of applied filters


ejgullo:
screen shot 2017-09-11 at 11 30 55 am

screen shot 2017-09-11 at 11 33 16 am

Seeing some weirdness with an empty set loading in, are we checking for duplicates/empty child values on load?

ejgullo:
Moving forward; issue identified as a result of the data that is passed in.  Any changes will need to happen on backend, if at all.

PR #402, opened 2017-09-19
Download Modal

kevinli-work:
- [x] API integration testing
- [x] Code reviewed
- [x] API merged into dev: https://github.com/fedspendingtransparency/usaspending-api/pull/541
- [x] Redirect to bulk download when more than 500k rows are requested 
- [x] Download row count API is implemented fedspendingtransparency/usaspending-api#606
- [x] Warn user about closing window before download is initiated

* Implements the download button
* Need to validate file downloading behavior on file completion
* Need to validate integration with transaction generation

* No redirection to bulk download (doesn't exist), download button is disabled when 500k rows exceeded


src/_scss/pages/modals/fullDownload/_filterBar.scss
```diff
@@ -0,0 +1,50 @@
...
+
+            font-size: $smallest-font-size;
+
+            &:hover, &:active {
+                background-color: $color-white;
+            }
+
+            .filter-item-title {
+                color: $color-gray;
+            }
```
ebdabbs: Missing `.filter-item-remove-container` styling for the x icon
filter-item

tests/redux/reducers/search/downloadReducer-test.js
```diff
@@ -0,0 +1,104 @@
+/**
+ * downloadReducer-test.js
+ * Created by Emily Gullo 08/24/2017
+ */
+
+import downloadReducer from 'redux/reducers/search/downloadReducer';
+import { List } from 'immutable';
+
+const initialState = {
```
ebdabbs: This can be imported from downloadReducer:
`import downloadReducer, { initialState } from 'redux/reducers/search/downloadReducer';`

ejgullo: 👍

src/_scss/pages/modals/fullDownload/_filterBar.scss
ejgullo: This was an issue with the compressed version of the tags that should be displaying.  Contract filters (Type of Pricing,  Extent Competed, Set-Aside) were missing this boolean - updated!  Now no tags should have the option to remove in the download modal.

PR #412, opened 2017-09-22
Writing new tests for:

ejgullo:
- AccountAwardsHeaderCellContainer
- AgencyFooterContainer
- AgencyLandingHeaderCellContainer
- AwardTypeContainer
- LocationSearchContainer

### DIFF
--- a/tests/containers/agency/AgencyFooterContainer-test.jsx
+++ b/tests/containers/agency/AgencyFooterContainer-test.jsx
@@ -10,10 +10,10 @@ import sinon from 'sinon';
 import { AgencyFooterContainer } from
     'containers/agency/AgencyFooterContainer';
 import Router from 'containers/router/Router';
-import * as AgencyHelper from 'helpers/agencyHelper';
 
-import { mockLoad, mockAgencyApi } from './mockAgency';
+import { mockCgacLoad } from './mockAgency';
 
+jest.mock('helpers/agencyHelper', () => require('./agencyHelper'));
 // force Jest to use native Node promises
 // see: https://facebook.github.io/jest/docs/troubleshooting.html#unresolved-promises
 global.Promise = require.requireActual('promise');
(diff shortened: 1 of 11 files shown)


tests/containers/agency/AgencyFooterContainer-test.jsx
```diff
@@ -0,0 +1,58 @@
...
+global.Promise = require.requireActual('promise');
+
+describe('AgencyFooterContainer', () => {
+    describe('loadAgency', () => {
+        it('should set the state of the loaded agency', async () => {
+            const container = mount(<AgencyFooterContainer
+                clearAllFilters={jest.fn()}
+                updateSelectedAwardingAgencies={jest.fn()} />);
+
+            container.instance().loadAgency('123');
```
kevinli-work: `loadAgency` calls `AgencyHelper`, which is not mocked. As a result, this makes a real API call. Unit tests are not integration tests and should have a scope restricted only to the target unit (`AgencyFooterContainer`) - we should assume that all the inputs are correct (they would be tested separately). Additionally, if the API data changes (such as the DB IDs change), the tests could fail even if the code itself is still valid.

Please make sure you are mocking the API calls. A good way to validate you are mocking data is to use clearly fake data (make up a fake agency) rather than basing the mock data in `mockAgency` on real API responses. Because the API will never return this fake data, you will know definitively if your test mocks are in fact mocks or if they are calling the real API.

tests/containers/search/filters/location/LocationSearchContainer-test.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+const toggleValue = {
+    target: {
+        value: 'foreign'
+    }
+};
+
+describe('LocationSearchContainer', () => {
+    describe('selectLocation', () => {
+        it('should update the location name in redux', () => {
```
kevinli-work: What about when the location is not valid or no location is provided? Tests should cover as many cases as possible so we can identify when the code breaks on these cases (and on which cases it broke).

tests/containers/agency/AgencyFooterContainer-test.jsx
```diff
@@ -10,33 +10,63 @@ import sinon from 'sinon';
...
+import { mockLoad, mockAgencyApi } from './mockAgency';
 
 // force Jest to use native Node promises
 // see: https://facebook.github.io/jest/docs/troubleshooting.html#unresolved-promises
 global.Promise = require.requireActual('promise');
 
+
+const routerPushSpy = sinon.spy(Router.history, 'push');
+
+const mockAgencyHelper = (functionName, event, expectedResponse) => {
```
kevinli-work: We have stopped doing this type of helper mock several sprints ago. Please review how they are being mocked here: https://github.com/fedspendingtransparency/usaspending-website/blob/dev/tests/containers/agency/visualizations/FederalAccountContainer-test.jsx#L14

tests/containers/agency/AgencyFooterContainer-test.jsx
```diff
@@ -10,33 +10,63 @@ import sinon from 'sinon';
...
 describe('AgencyFooterContainer', () => {
     describe('loadAgency', () => {
         it('should set the state of the loaded agency', async () => {
+            mockAgencyHelper('fetchAgencyCgacCode', 'resolve', mockAgencyApi);
             const container = mount(<AgencyFooterContainer
                 clearAllFilters={jest.fn()}
                 updateSelectedAwardingAgencies={jest.fn()} />);
 
-            container.instance().loadAgency('123');
+            container.instance().loadAgency();
```
kevinli-work: You should still pass an argument into the the function...

PR #420, opened 2017-09-25
Spending Explorer Updates

ebdabbs:
- Prevents treemap cells from getting cut off when there are negative obligations
- Shows a message when a user clicks on an option and there are no associated awards at the next step/ total is $0
- Closes the dropdown when user clicks outside it

### DIFF
--- a/src/js/components/explorer/detail/NoAwardsScreen.jsx
+++ b/src/js/components/explorer/detail/NoAwardsScreen.jsx
@@ -1,0 +1,45 @@
+/**
+ * NoAwardsScreen.jsx
+ * Created by Lizzie Salita 9/21/17
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import * as Icons from 'components/sharedComponents/icons/Icons';
+
+const propTypes = {
+    rewindToFilter: PropTypes.func,
+    currentIndex: PropTypes.number
+};
+
+export default class NoAwardsScreen extends React.Component {
+    constructor(props) {
+        super(props);
+        this.clickedLink = this.clickedLink.bind(this);
+    }
+
+    clickedLink() {
+        const previousIndex = this.props.currentIndex - 1;
+        console.log(previousIndex);
+        this.props.rewindToFilter(previousIndex);
+    };
+
+    render() {
+        return (
+            <div className="explorer-no-awards">
+                <div className="no-awards-message">
+                    <span className="info-icon-circle">
+                        <Icons.InfoCircle />
+                    </span>
+                    You&#8217;ve reached a point with no associated awards.
+                </div>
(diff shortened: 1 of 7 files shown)


src/js/components/explorer/detail/visualization/treemap/ExplorerTreemap.jsx
```diff
@@ -52,6 +53,10 @@ export default class ExplorerTreemap extends React.Component {
 
     buildVirtualChart(props) {
         const data = props.data.toJS();
+
+        // remove the negative values from the data because they can't be displayed in the treemap
+        remove(data, (v) => v.amount <= 0);
```
kevinli-work: We need to add additional logic around handling an empty `data` set. If the full `data` array is negative or 0, then data becomes an empty array. However, by passing the empty array through the d3 `hierarchy` and `tree().leaves()` functions, we still end up with a single node in `treeItems`.

Instead of checking only if `treeItems.length` is 0, we should also check if `data.length` is 0 and set an empty `virtualChart` array in state.

src/js/components/explorer/detail/visualization/toolbar/BreakdownDropdown.jsx
```diff
@@ -34,12 +34,18 @@ export default class BreakdownDropdown extends React.Component {
...
+        this.setWrapperRef = this.setWrapperRef.bind(this);
+        this.handleClickOutside = this.handleClickOutside.bind(this);
     }
 
     componentWillMount() {
         this.prepareOptions(this.props);
     }
 
+    componentDidMount() {
+        document.addEventListener('mousedown', this.handleClickOutside);
```
kevinli-work: Instead of always listening for this event (which could add to our overhead), what if we only added the listener when the dropdown is expanded (we have access to the `expanded` state here) and then removed when the dropdown collapses again and/or the component unmounts?

src/js/components/explorer/detail/NoAwardsScreen.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+export default class NoAwardsScreen extends React.Component {
+    constructor(props) {
+        super(props);
+        this.clickedLink = this.clickedLink.bind(this);
+    }
+
+    clickedLink() {
+        const previousIndex = this.props.currentIndex - 1;
+        console.log(previousIndex);
```
kevinli-work: Can we remove this console.log

src/js/components/explorer/detail/visualization/treemap/ExplorerTreemap.jsx
ebdabbs: Changed the check to `if (treeItems.length === 0 || data.length === 0)`

PR #451, opened 2017-10-10
Update sass speedup

ebdabbs:
- Each page `require`s its own Sass file instead of all Sass files being imported by `custom.scss`
- Reduces compile time for a style change from ~17s to ~5s


kevinli-work:
This still builds a large CSS file that is `<link>`ed to in our HTML head. If we're going to split out the Sass and dynamically inject styles, should we consider eliminating the CSS file entirely?

Take a look at the `index.html` file as well as the `ExtractTextPlugin` extractions in the Webpack config

kevinli-work:
I have convinced myself this is good

PR #466, opened 2017-10-17
Feature Explorer Table View

ebdabbs:
Implements the table view in the Spending Explorer
- Toggle between treemap and table view
- Sort results by column
- Show pagination when there are more than 20 results


src/js/containers/explorer/detail/table/ExplorerTableContainer.jsx
```diff
@@ -0,0 +1,259 @@
...
+    componentWillReceiveProps(nextProps) {
+        if (nextProps.results !== this.props.results) {
+            this.parseResults(nextProps.results);
+        }
+    }
+
+    componentDidUpdate(prevProps) {
+        if (this.props.order !== prevProps.order) {
+            // table sort changed
+            this.orderResults();
```
kevinli-work: When does `orderResults` get called when the initial data set is loaded? It only seems to be sorted once the order prop changes.

src/js/containers/explorer/detail/table/ExplorerTableContainer.jsx
```diff
@@ -0,0 +1,259 @@
...
+
+    onChangePage(pageNumber, inRange) {
+        // Change page number in Redux state
+        if (inRange) {
+            this.props.setExplorerTablePage(pageNumber);
+        }
+    }
+
+    getPager(totalItems, currentPage, pageSize) {
+        // calculate total pages
```
kevinli-work: The logic seems like it should belong in the pagination component.

src/js/containers/explorer/detail/table/ExplorerTableContainer.jsx
```diff
@@ -0,0 +1,259 @@
...
+
+            const column = {
+                columnName: col,
+                displayName,
+                defaultDirection: sortOrder[col]
+            };
+            columns.push(column);
+        });
+
+        this.setState({
```
kevinli-work: It looks like multiple chains of `setState` are being called via these callbacks. Every time state is updated, the component and all its child components re-render (or at least diff). If we need all these operations (showColumns, parseData, setPages) to be done before a meaningful render can occur, it is better to perform all these operations and then do a single setState to reduce unusable render passes.

src/js/containers/explorer/detail/table/ExplorerTableContainer.jsx
```diff
@@ -0,0 +1,259 @@
...
+
+        // Get new pager object for specified page
+        const pager = this.getPager(results.length, page, 20);
+
+        // Get new page of items from results
+        const pageOfItems = results.slice(pager.startIndex, pager.endIndex + 1);
+
+        // update state
+        this.setState({
+            pager,
```
kevinli-work: Avoid storing React components in state. React evaluates inbound state/props for changes from the current set of state/props to determine if it needs to re-render. However, if components are stored in state, a comparison can't be made so a render happens by default. A better approach is to precalculate a complex child component's data and store it as an object that the child component can parse, compare, and display.

src/js/components/sharedComponents/Pagination.jsx
```diff
@@ -0,0 +1,61 @@
...
+                </div>
+                <ul className="pager">
+                    <li className={pager.currentPage === 1 ? 'disabled' : ''}>
+                        <button onClick={() => this.setPage(pager.currentPage - 1)}>{`<`}</button>
+                    </li>
+                    {pager.firstButton}
+                    {pager.prevEllipses}
+                    {pager.pages.map((page, index) =>
+                        <li key={index} className={pager.currentPage === page ? 'active' : ''}>
+                            <button onClick={() => this.setPage(page)}>{page}</button>
```
kevinli-work: Avoid defining new functions in the `render` return whenever possible. This is because the browser will instantiate a new instance of the function each time the component re-renders.

Instead, bind to an existing function:

```
constructor(props) {
    super(props);
    this.myFunction = this.myFunction.bind(this);
    // this sets the context to this component instance, so things like this.props or this.state will work correctly
}

myFunction() {
     // something, the actual function
}

render() {
    // this calls the this.myFunction you set in the constructor
    return (
        <button onClick={this.myFunction} />
    );
}
```

src/js/components/explorer/detail/visualization/table/cells/LinkCell.jsx
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    clickedLink() {
+        this.props.selectedRow(this.props.id, this.props.data);
+    }
+
+    render() {
+        const name = this.props.name;
+        // highlight the matched string if applicable
+        // if (this.props.agencySearchString !== '') {
```
kevinli-work: Delete if unused

src/_scss/pages/explorer/detail/visualization/table/cells/_explorerHeaderCell.scss
```diff
@@ -0,0 +1,49 @@
+@mixin explorerHeaderCell() {
```
kevinli-work: Why is this a mixin?

geoffgeiling:
Design review - approved

PR #481, opened 2017-10-23
Explain Why Download is Disabled

kevinli-work:
* Adds a hover tooltip to explain why download is not available when there are more than 500,000 rows
* Adds an exclamation icon to the download button when it is not available

### DIFF
--- a/src/_scss/layouts/search/header/_downloadButton.scss
+++ b/src/_scss/layouts/search/header/_downloadButton.scss
@@ -1,34 +1,107 @@
 .download-wrap {
     position: relative;
 
+    .download-hover {
+        position: absolute;
+        top: rem(-23);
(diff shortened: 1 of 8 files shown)


src/_scss/layouts/search/header/_downloadButton.scss
```diff
@@ -1,34 +1,111 @@
...
+
+        .label {
+            @include flex(1 1 auto);
             color: $color-white;
-            border: 1px solid $color-white;
+            text-transform: uppercase;
+            font-size: 1em;
+            font-weight: $font-semibold;
+            letter-spacing: 0.075em;
+        }
```
ebdabbs: Add `text-align: center` for even spacing when there is no warning icon
download-button

PR #509, opened 2017-11-13
Recipient Filter v2.0

mab6bf:
- Restructures the Recipient filter to perform in a similar manner to the Keyword filter
- Automatically populates the filter with a valid Recipient search response
- Disables the input field and Search button during the API call to indicate "something is happening" and prevent the ability to DDoS the API through repeated requests
- Increases the minimum length of the Recipient search string from 2 to 3 characters

https://federal-spending-transparency.atlassian.net/browse/DS-1882

- [x] Code Review
- [x] Design Review


kevinli-work:
Why is this an autocomplete? The story and the design call for this filter to use a submit button, the implication being that API calls are only performed when the button/enter is pressed rather than on each keystroke.

kevinli-work:
screen shot 2017-11-13 at 7 02 51 pm

src/js/containers/search/filters/recipient/RecipientNameDUNSContainer.jsx
```diff
@@ -30,178 +25,109 @@ export class RecipientNameDUNSContainer extends React.Component {
...
+                    const autocompleteData = res.data.results;
+
+                    this.setState({
+                        inFlight: false
+                    });
+
+                    // Show error if there are no recipient IDs
+                    if (autocompleteData.recipient_id_list.length === 0) {
+                        this.setState({
+                            showWarning: true
```
kevinli-work: Can you re-enable the submit button on no results? Otherwise the user can't search again.

src/js/containers/search/filters/recipient/RecipientNameDUNSContainer.jsx
```diff
@@ -30,178 +25,109 @@ export class RecipientNameDUNSContainer extends React.Component {
...
+                        });
+                    }
+                    // Prevent duplicate results
+                    else if (!this.props.selectedRecipients.has(autocompleteData.search_text)) {
```
kevinli-work: Why not do this check before we send out the API call (since we know both the redux contents and the `search_text` value locally)?

src/js/models/search/SearchAwardsOperation.js
```diff
@@ -145,8 +146,15 @@ class SearchAwardsOperation {
 
         // Add Recipients, Recipient Scope, Recipient Locations, and Recipient Types
         if (this.selectedRecipients.length > 0) {
-            filters[rootKeys.recipients] = this.selectedRecipients.map(
-                (recipient) => recipient.legal_entity_id);
+            let recipients = [];
+
+            this.selectedRecipients.forEach((recipient) => {
+                const tempRecipients = recipients;
```
kevinli-work: Do we need `tempRecipients`? `concat` should return a new array

src/js/containers/search/filters/recipient/RecipientNameDUNSContainer.jsx
kevinli-work: Nvm, this seems to only occur when the server errors out.

mab6bf: Yeah, forgot to update `inFlight` in the `catch`

src/js/containers/search/filters/recipient/RecipientNameDUNSContainer.jsx
mab6bf: Good call

src/js/models/search/SearchAwardsOperation.js
mab6bf: Done

geoffgeiling:
Design Reivew Complete based on screenshots with Mike Bray. I will need to test that the filters are filtering properly in staging once it's pushed.

PR #531, opened 2017-11-21
New About Page

mab6bf:
- Implements new About page
- Does not yet include the "View Agency Quarterly Submission Status" button, as the Quarterly Submission Status page hasn't been built yet

Mock: https://projects.invisionapp.com/d/main#/console/8586929/262682080/preview
Story: https://federal-spending-transparency.atlassian.net/browse/DS-1944

- [x] Code Review
- [x] Design Review
- [x] Content Approval


src/_scss/pages/about/_section.scss
```diff
@@ -0,0 +1,70 @@
...
+                    display: block;
+                    font-weight: $font-semibold;
+                }
+            }
+        }
+
+        button {
+            display: block;
+            margin: rem(19) 0 rem(38);
+            padding: rem(16) rem(32) rem(13);
```
kevinli-work: Any reason why the button top padding is different from the bottom padding? It results in non-vertically centered text.

src/_scss/pages/about/_contentWrap.scss
```diff
@@ -0,0 +1,18 @@
...
+    color: $color-base;
+    border-top: 1px solid $color-gray-border;
+    border-right: 1px solid $color-gray-border;
+    border-bottom: 1px solid $color-gray-border;
+
+    .about-padded-content {
+        padding: rem(35);
+    }
+
+    @media(min-width: $large-screen) {
```
kevinli-work: Please use the standard media mixin we're using everywhere else on the site `@include media($screen-var)`

src/_scss/pages/about/header/header.scss
```diff
@@ -0,0 +1 @@
+@import "components/pageTitleBar/pageTitleBar";
```
kevinli-work: Why do this? Why not just do this import in the parent file and delete this file?

src/js/components/about/AboutSidebar.jsx
```diff
@@ -0,0 +1,137 @@
+/**
```
kevinli-work: This seems almost identical to AgencySidebar. Any reason why we couldn't just use that component and make the modifications necessary to accept CSS class name as a prop (or wrap around child content)?

src/_scss/pages/about/header/header.scss
mab6bf: Copied CSS from Agency - will remove

src/_scss/pages/about/_section.scss
mab6bf: Was matching to the mock via Inspect mode. Assuming the mock wasn't fully fleshed out, I can just remove that last `rem(13)`.

src/_scss/pages/about/_contentWrap.scss
mab6bf: Done

src/js/components/about/AboutSidebar.jsx
mab6bf: Done

PR #537, opened 2017-11-21
Configurable Beta Banners

kevinli-work:
* When the `IN_BETA` flag is missing or set to `false` in `GlobalConstants_[env].js`, the beta banner will be replaced by a post-beta banner and the "beta" tag will disappear from the logo

https://federal-spending-transparency.atlassian.net/browse/DS-1900

- [x] Code review
- [x] Design review


AGuyNamedMarco:
Looking solid! Think we could reduce the padding even more?  I see it's at 1.5rem for top and bottom... could .5rem work?

ebdabbs:
Updates address https://federal-spending-transparency.atlassian.net/browse/DS-2005

PR #542, opened 2017-11-22
Performance Option: Fewer Rows

kevinli-work:
* Award search table requests 30 rows at a time instead of 60
* Can optionally be merged *in addition* to column changes


mab6bf:
Not going with this one, as it won't noticeably impact performance

PR #555, opened 2017-11-28
Update kerning of logo

AGuyNamedMarco:
Update logo images with proper kerning.


src/_scss/layouts/default/header/_logo.scss
```diff
@@ -42,19 +42,19 @@
...
             @include border-radius(2px);
-            @include media(360px) {
-                left: rem(228);
+            @include media(359px) {
+                left: rem(223);
             }
             @include media($large-screen) {
                 font-size: rem(10);
-                left: rem(280);
+                left: rem(274);
```
kevinli-work: Please make sure you've also accounted for the "beta" label positioning on the home page (you may need to override in `pages/homepage/header/header.scss`

You can see the source of each CSS property if you run `npm run sass` and look at the Chrome devtools element inspector.

kevinli-work: screen shot 2017-11-30 at 12 55 24 pm

AGuyNamedMarco: The last commit should fix the homepage 'beta' label

PR #562, opened 2017-12-01
Center map loading message

kevinli-work:
* Move loading message to center of the map


geoffgeiling:
Design Review complete

PR #568, opened 2017-12-05
Revert "Fix: Allow forced overrides of map measurement diff"

mab6bf:
Reverts fedspendingtransparency/usaspending-website#567


kevinli-work:
Unnecessary

PR #571, opened 2017-12-08
Submit-based Award ID

kevinli-work:
* Convert Award ID filter from autocomplete to submit-based search
* Only allow a single award ID filter to be applied at a time
* Standardize individual filter submit buttons to use the magnifying glass design

https://federal-spending-transparency.atlassian.net/browse/DS-1989

- [ ] Code review
- [x] fedspendingtransparency/usaspending-api#796 is merged


src/_scss/pages/search/filters/awardAmount/_specificAwardAmount.scss
```diff
@@ -26,6 +28,10 @@
             }
         }
 
+        .award-amount-submit {
+            @include singleSubmit;
```
ebdabbs: specificawardamount
Icon goes off-center between mobile and 1342 px

src/js/components/search/filters/awardID/AwardIDSearch.jsx
```diff
@@ -15,6 +15,28 @@ const propTypes = {
...
+
+    changedInput(e) {
+        this.setState({
+            awardID: e.target.value
+        });
+    }
+
+    applyAwardID(e) {
+        e.preventDefault();
+        this.props.toggleAwardID(this.state.awardID);
```
ebdabbs: Reset `awardID`in the state after the filter is applied to clear the input value

PR #579, opened 2017-12-19
Fix spending time graph

kimminahm:
[https://federal-spending-transparency.atlassian.net/browse/DS-1790](Jira)

- [ ] Code Review
- [x] Design Review

PR Features
- Fixes tooltip flipping when it reaches the edge of the chart
- Variable padding between bars so that we never get negative width
- Labels for all views are by year, on hover only shows the label for the hovered bar
- Removes bar components from state and creates in render function


geoffgeiling:
Design review is complete. There are a few small things I think we'll want to adjust but we can do it at a later date (i.e. the tooltip covering the text on the x-axis). They will probably require some design adjustments.

PR #587, opened 2017-12-21
Update Subtiers Based on Names

ebdabbs:
- [ ] [Backed PR](https://github.com/fedspendingtransparency/usaspending-api/pull/830) merged

- [x] Code review

- Removes the `subtier_agency_id` field, which will no longer be returned by the `/list_agencies` endpoint

- Passes the subtier agency name as a string instead of its id to the `/bulk_download/awards` endpoint

https://federal-spending-transparency.atlassian.net/browse/DS-2065

### DIFF
--- a/src/js/components/bulkDownload/awards/filters/AgencyFilter.jsx
+++ b/src/js/components/bulkDownload/awards/filters/AgencyFilter.jsx
@@ -70,8 +70,7 @@ export default class AgencyFilter extends React.Component {
         e.preventDefault();
         const target = e.target;
         this.props.updateFilter('subAgency', {
-            id: target.value,
-            name: target.name
+            name: target.value
         });
 
         this.setState({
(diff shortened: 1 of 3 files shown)


src/js/components/bulkDownload/awards/filters/AgencyFilter.jsx
```diff
@@ -129,16 +128,15 @@ export default class AgencyFilter extends React.Component {
         ));
 
         // Create the sub-agency options
-        const subAgencies = this.props.subAgencies.map((subAgency) => (
+        const subAgencies = this.props.subAgencies.map((subAgency, index) => (
             <li
                 className="field-item"
-                key={`field-${subAgency.subtier_agency_id}`}>
+                key={`field-${index}`}>
```
kevinli-work: Are subtier agency names unique? If so, can you use that instead of `index` as the key?

PR #591, opened 2017-12-28
Filter Submit Button

kevinli-work:
https://federal-spending-transparency.atlassian.net/browse/DS-1983

* Rearchitects award search page to support submit button on filter bar
* Implements submit button on filter bar
* Implements default blank state
* Implements loading, no result, and server error message screens
* Locks submit button and tabs from being interactive when API requests are in flight
* Adds tag previews to Date Range and Keyword filters
* Updated top filter bar design and is now non-interactive

* Does **not** implement submit button hints - that is coming in a subsequent PR

- [ ] Code review
- [x] Design review


AGuyNamedMarco:
Design is looking solid.  Thanks Kevin!

src/js/components/search/SearchResults.jsx
```diff
@@ -18,6 +18,8 @@ const propTypes = {
     isMobile: PropTypes.bool,
     filterCount: PropTypes.number,
     showMobileFilters: PropTypes.bool,
+    requestsComplete: PropTypes.bool,
+    noFiltersApplied: PropTypes.bool,
     toggleMobileFilters: PropTypes.func,
     clearAllFilters: PropTypes.func
```
ebdabbs: Mobile "Clear all filters" button is not working because this prop is no longer passed down from `SearchPage`. Can that button be removed since we have the reset button?

ebdabbs: Mobile "Clear all filters" button is not working because this prop is no longer passed down from `SearchPage`. Can that button be removed since we have the reset button?

kevinli-work: Removed

PR #594, opened 2018-01-02
Feature download experience

kimminahm:
https://federal-spending-transparency.atlassian.net/browse/DS-2054
- Matches the download experience for agency download and bulk download
- Adding auto link copy for download experience
- Updating language in the download modals as well as the bottom bar for both pages
- Download links added to the bottom bar 
- [ ] Code Review
- [x] Design Review


src/js/containers/search/modals/fullDownload/DownloadBottomBarContainer.jsx
```diff
@@ -108,6 +109,7 @@ export class DownloadBottomBarContainer extends React.Component {
         this.request.promise
             .then((res) => {
                 this.props.setDownloadExpectedFile(res.data.file_name);
+                this.props.setDownloadExpectedUrl(res.data.url);
                 this.checkStatus();
```
kevinli-work: In the `downloadFile()` method of this component, can you ensure that it closes the download modal if it is open when the download occurs?

Currently, the Redux store is reset, so the modal stays open but clears itself.

screen shot 2018-01-02 at 11 56 35 am

kimminahm: auto closes when download completes

kimminahm:
Gita cannot check the box but she approves design

PR #608, opened 2018-01-10
adding padding to keep data from being blocked

kimminahm:
https://federal-spending-transparency.atlassian.net/browse/DEV-107

- Increasing row height padding onto bottom when there is only one row so that the horizontal scroll bar doesnt cover it up

- [x] Design Review
- [x] Code Review


kimminahm:
yes

geoffgeiling:
can we increase that padding a bit more? I know it's not covered now but still feels weird with no extra white space right after the text

PR #613, opened 2018-01-16
Fix Bulk Download Date Reset

kimminahm:
https://federal-spending-transparency.atlassian.net/browse/DEV-87
- Fixes bug where the date range would not reset when the bulk download form was reset so you could not increase the date range


src/js/components/sharedComponents/DatePicker.jsx
```diff
@@ -229,6 +229,10 @@ export default class DatePicker extends React.Component {
             });
         }
 
+        else if (!this.props.value) {
+            disabledDays.length = 0;
```
kevinli-work: Why are you changing the length of the array instead of the array itself?

src/js/components/bulkDownload/awards/filters/dateRange/TimePeriodFilter.jsx
```diff
@@ -97,6 +97,11 @@ export default class TimePeriodFilter extends React.Component {
                 newState.endDateUI = null;
             }
         }
+
+        if ((nextProps.filterTimePeriodEnd === '' && nextProps.filterTimePeriodStart === '') && nextProps.valid) {
```
kevinli-work: Is this necessary? Doesn't the parent `AwardDataContent` already run `validateForm` on `componentWillReceiveProps` anyway?

src/js/components/sharedComponents/DatePicker.jsx
kimminahm: Since disabled days is a const I can either update the length property to 0 which will reset it to an empty array or splice it onto itself at position 0 or iteratively pop off each element. I thought this was the most straightforward way to do it.

src/js/components/bulkDownload/awards/filters/dateRange/TimePeriodFilter.jsx
kimminahm: it will run `validateForm` but when you hit reset it doesnt actually set the `validDate` state to false so it will allow you to submit a request without selecting a date on the second attempt

PR #614, opened 2018-01-16
Fix - Federal Account Table Columns

mab6bf:
Does not need to go in this deploy.

- Columns in the Federal Account table now match those in the Award Search table for each Award Type

Bug: https://federal-spending-transparency.atlassian.net/browse/DEV-116

Note: For full testing, @michaeldhess is looking into finding a Federal Account in the dev database that has corresponding awards to display in the table. However, even without result data in the table (which allows the user to scroll through the column headers), the visible column headers can be seen updating on each tab change.

Use 2421

### DIFF
--- a/src/js/components/account/awards/LegacyResultsTable.jsx
+++ b/src/js/components/account/awards/LegacyResultsTable.jsx
@@ -13,9 +13,9 @@ import IBTable from 'components/sharedComponents/IBTable/IBTable';
 import ResultsTableGenericCell from 'components/search/table/cells/ResultsTableGenericCell';
 import ResultsTableAwardIdCell from 'components/search/table/cells/ResultsTableAwardIdCell';
 
-import LegacyTableHeaderCell from './LegacyTableHeaderCell';
-
 import AccountTableSearchFields from 'dataMapping/search/accountTableSearchFields';
+
+import LegacyTableHeaderCell from './LegacyTableHeaderCell';
 
 const propTypes = {
     results: PropTypes.array,
--- a/src/js/components/account/awards/LegacyResultsTable.jsx
+++ b/src/js/components/account/awards/LegacyResultsTable.jsx
@@ -15,7 +15,7 @@ import ResultsTableAwardIdCell from 'components/search/table/cells/ResultsTableA
 
 import LegacyTableHeaderCell from './LegacyTableHeaderCell';
 
-import AccountTableSearchFields from '../../../dataMapping/search/accountTableSearchFields';
+import AccountTableSearchFields from 'dataMapping/search/accountTableSearchFields';
 
 const propTypes = {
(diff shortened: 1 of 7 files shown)


kevinli-work:
Pending getting a federal account to test this with.

kevinli-work:
* Award count tabs are all 0 (if we don't have the counts, can you hide the count badge?)
* Awarding Agency/Subagency doesn't appear:
screen shot 2018-01-19 at 12 58 14 pm

* Loans tab causes a 500 error: `Cannot resolve keyword 'face_value_loan_guarantee' into field`

kevinli-work:
* Sorting by any field other than award amount returns 0 results

src/js/components/account/awards/LegacyResultsTable.jsx
```diff
@@ -8,16 +8,15 @@
...
-import { formatMoney } from 'helpers/moneyFormatter';
-
 import IBTable from 'components/sharedComponents/IBTable/IBTable';
 
 import ResultsTableGenericCell from 'components/search/table/cells/ResultsTableGenericCell';
 import ResultsTableAwardIdCell from 'components/search/table/cells/ResultsTableAwardIdCell';
 
 import LegacyTableHeaderCell from './LegacyTableHeaderCell';
 
+import AccountTableSearchFields from '../../../dataMapping/search/accountTableSearchFields';
```
kevinli-work: o rly? You should be able to use the module path from `src/js/`

src/js/containers/account/awards/AccountAwardsContainer.jsx
```diff
@@ -11,15 +11,16 @@ import { uniqueId } from 'lodash';
...
+import AccountTableSearchFields from 'dataMapping/search/accountTableSearchFields';
 import { awardTypeGroups } from 'dataMapping/search/awardType';
 import { awardTableColumnTypes } from 'dataMapping/search/awardTableColumnTypes';
 import * as SearchHelper from 'helpers/searchHelper';
 
 import AccountAwardSearchOperation from 'models/account/queries/AccountAwardSearchOperation';
-
 import AccountAwardsSection from 'components/account/awards/AccountAwardsSection';
 
+import BaseFederalAccountAwardRow from '../../../models/v2/BaseFederalAccountAwardRow';
```
kevinli-work: See above comment about module paths

src/js/components/account/awards/LegacyResultsTable.jsx
mab6bf: Fixed

src/js/containers/account/awards/AccountAwardsContainer.jsx
mab6bf: Fixed

PR #617, opened 2018-01-18
Update keyword search

ebdabbs:
https://federal-spending-transparency.atlassian.net/browse/DEV-323

- Implements tab counts and selecting a default tab
- Adds a loading state for the search summary
- Links Award ID column to award profile pages
- Allows sorting on all columns
- Checks for the current sort field when switching tabs
-  Since Elasticsearch currently only has data from FY 2015, 2016, & 2017, limits the search summary request to those years

Checklist
- [x] `v2/search/spending_by_transaction_count/` endpoint merged: https://github.com/fedspendingtransparency/usaspending-api/pull/899
- [x] Tab counts API integration, [gist here](https://hackmd.io/AwRg7ATAzAnArMAtAYwnMiAskBsiAc0miAJvjCMlApnDjEA=)
- [x] Sorting on Mod, Awarding Sub Agency, Loan Value, and Subsidy Cost fixed on the backend
- [x] Code review
- [x] Design review


src/js/containers/keyword/table/ResultsTableContainer.jsx
```diff
@@ -0,0 +1,325 @@
...
+            });
+    }
+
+    parseTabCounts(data) {
+        const transactionCounts = data.results;
+        let firstAvailable = '';
+        let i = 0;
+
+        // Set the first available award type to the first non-zero entry in the
+        while (firstAvailable === '' && i < tableTypes.length) {
```
kevinli-work: It would be my preference to use a `for` and `break;` instead of a `while` in order to minimize the risk of an infinite loop.

src/js/components/keyword/table/ResultsTableSection.jsx
```diff
@@ -65,7 +66,7 @@ export default class ResultsTableSection extends React.Component {
                 visibleWidth={this.state.tableWidth} />
         );
 
-        if (!this.props.keywordApplied) {
+        if (this.props.keyword === '') {
```
kevinli-work: Since `''` is a [falsy value](http://www.ecma-international.org/ecma-262/6.0/#table-10) and the `if` will coerce it to a boolean, you can actually just do `if (this.props.keyword)`.

src/js/containers/keyword/table/ResultsTableContainer.jsx
```diff
@@ -0,0 +1,323 @@
...
+
+            if (transactionCounts[tableType] > 0) {
+                firstAvailable = tableType;
+                break;
+            }
+        }
+
+        // If none of the award types are populated, set the first available tab to be the
+        // first tab in the table
+        if (firstAvailable === '') {
```
kevinli-work: Can you change this equality check to a falsy string check (`if (!firstAvailable)`)? There are also a few other places this happens in this container.

src/js/components/keyword/header/KeywordHeader.jsx
```diff
@@ -30,19 +30,25 @@ export class KeywordHeader extends React.Component {
             const primeCountUnits = MoneyFormatter.calculateUnitForSingleValue(primeCount);
             const primeAmountUnits = MoneyFormatter.calculateUnitForSingleValue(primeAmount);
 
-            let primeCountPrecision = 0;
-            if (primeCountUnits.unit !== 1) {
-                primeCountPrecision = 1;
+            if (primeCountUnits.unit >= 1000000) {
```
kevinli-work: To avoid future (or current) typos where we add/drop a zero by accident, can you use the `unitValues` in `moneyFormatter` for this?

PR #631, opened 2018-01-29
New Homepage: Phase 5 - Wrap Up

kevinli-work:
https://federal-spending-transparency.atlassian.net/browse/DEV-331

* Delete old home page files
* Promote new home page to the default landing route
* Agency pages recycled treemap styling from old home page; those files have been moved to the agency page

- [x] Design review
- [x] Code review


src/_scss/pages/homepage/hero/hero.scss
```diff
@@ -50,7 +50,7 @@
         }
 
         strong.homepage-hero__headline_weight_bold {
```
mab6bf: Should we rename this class now?

kevinli-work: I don't think so because it is still relatively bold

PR #643, opened 2018-02-02
Fix - Download Center (Bulk Download) Restrictions

mab6bf:
- Replaces all instances of "Bulk Download" with "Download Center"
- Removes the "All Time" option in Download Center
- Limits Download Center functionality to a year's worth of data
- Fixes lingering bug with Date Pickers failing to open to a monthly view with the selected date in non-current years
- Adds a generic Sticky Header that replaces Search Header and Keyword Header
- Implements Sticky Header in Download Center
- Converts Sticky Header CSS to BEM
- Changes names of Award Data and Account Data to Custom Award Data and Custom Account Data, respectively
- Adds a new Award Data Archive link to the top of the Download Center sidebar
- Adds a "last 15 days" option

Issue: https://federal-spending-transparency.atlassian.net/browse/DEV-398

- [x] Code Review
- [x] Design Review


src/_scss/layouts/default/stickyHeader/header.scss
```diff
@@ -1,26 +1,26 @@
...
             position: fixed;
             top: 0;
             z-index: $z-header;
         }
     }
 
-    .search-header {
+    .sticky-header {
         @import "mixins/fullSectionWrap";
         @include fullSectionWrap(0, 0);
```
kevinli-work: We can't put `fullSectionWrap` at the block level because it is a mixin that primarily applies margin, which is [disallowed](https://en.bem.info/methodology/quick-start/#block).

src/_scss/layouts/default/stickyHeader/header.scss
```diff
@@ -1,26 +1,26 @@
...
+    &.sticky-header__container_sticky {
         @include media($medium-screen) {
             position: fixed;
             top: 0;
             z-index: $z-header;
         }
     }
 
-    .search-header {
+    .sticky-header {
```
kevinli-work: Why is the block defined inside an element member of the block itself?

src/_scss/pages/keyword/header/header.scss
```diff
@@ -1,39 +1,46 @@
-@import "layouts/tabbedSearch/header/header";
+@import "layouts/default/stickyHeader/header";
 
-.search-header-container {
-    .search-header {
-        .search-title {
+.sticky-header__container {
+    .sticky-header {
```
kevinli-work: The block should not be defined inside one of its member element.
Additionally, it seems like you are making a variant of the `sticky-header` block here for the keyword header bar. Since the `<StickyHeader />` component utilizes children to wrap around the header contents, maybe a better approach would be to make the keyword header bar's contents its own module, that handles the inner layout and formatting independently of the parent `sticky-header` block.

src/_scss/pages/keyword/header/header.scss
```diff
@@ -1,39 +1,46 @@
...
+            font-size: $small-font-size;
+
             @include media($medium-screen) {
                 @include span-columns(8);
                 @include display(flex);
             }
-            font-size: $small-font-size;
-            .summary-title {
+
+            .sticky-header__summary_title {
```
kevinli-work: These don't seem like true modifiers. For one, you should [not use modifiers alone](https://en.bem.info/methodology/quick-start/#a-modifier-cant-be-used-alone); basically a modifier class should provide only the styling necessary to change its parameter. Use composition to apply the primary layout/styling for the block or element:

```
<div className="keyword-header__summary keyword-header__summary_type_count">
```

src/js/components/keyword/KeywordPage.jsx
```diff
@@ -85,24 +87,85 @@ export default class KeywordPage extends React.Component {
...
+            }
+        }
+
+        return (
+            <div className="sticky-header__summary">
+                <div className="sticky-header__summary_title">
+                    Search Summary
+                </div>
+                <div className="sticky-header__summary_award-amounts">
+                    <div className="prime">
```
kevinli-work: Why are `prime` and `number` not BEM modifiers?

src/js/components/sharedComponents/DatePicker.jsx
```diff
@@ -227,6 +227,26 @@ export default class DatePicker extends React.Component {
                 before: this.props.opposite.toDate()
             });
         }
+        else if (this.props.type === 'startDateBulk' && this.props.opposite) {
```
kevinli-work: If `DatePicker` is a such a commonly used element across the site, does it make sense to hard-code a bulk download-specific ruleset into the shared component? Maybe we should modify the component so it can accept validation rules from parent components (via props) and put the bulk download rules in the bulk download code?

mab6bf: Yep. I moved the logic of `disabledDays` from here to the parents that implement `DatePicker`.

src/js/components/bulkDownload/awards/AwardDataContent.jsx
```diff
@@ -102,7 +102,7 @@ export default class AwardDataContent extends React.Component {
         return (
             <div className="download-data-content">
                 <div className="download-filters">
-                    <h2><span>Award Data</span> Download</h2>
+                    <h2><span>Custom Award Data</span></h2>
```
kevinli-work: This is `<span>` necessary?

src/js/components/bulkDownload/BulkDownloadPage.jsx
```diff
@@ -20,15 +21,21 @@ import BulkDownloadModalContainer from
 import BulkDownloadSidebar from './sidebar/BulkDownloadSidebar';
 
 export const dataTypes = [
+    {
```
kevinli-work: Can we get these from `dataMapping/navigation/menuOptions.js` to ensure that labeling and enable/disable state is centrally managed?

src/js/components/keyword/KeywordPage.jsx
```diff
@@ -85,24 +87,87 @@ export default class KeywordPage extends React.Component {
...
+        }
+
+        return (
+            <div className="keyword-header__summary">
+                <div className="keyword-header__summary-title">
+                    Search Summary
+                </div>
+                <div className="keyword-header__summary-award-amounts">
+                    <div className="keyword-header__summary-amount">
+                        Total Prime Award Amount: <span className="keyword-header__summary-amount_bold">{formattedPrimeAmount}</span>
```
kevinli-work: You can't use a modifier in isolation.

src/js/components/bulkDownload/BulkDownloadPage.jsx
mab6bf: Yep

src/js/components/keyword/KeywordPage.jsx
mab6bf: Fixed

PR #644, opened 2018-02-05
Fix Keyword Search Padding

ebdabbs:
- Adds more space between the search bar and results section
- Converts the search bar to BEM classes

- [x] Code review
- [x] Design review

### DIFF
--- a/src/_scss/pages/keyword/_searchBar.scss
+++ b/src/_scss/pages/keyword/_searchBar.scss
@@ -64,11 +64,11 @@
     .keyword-search-bar__info {
         font-size: $small-font-size;
         line-height: rem(26);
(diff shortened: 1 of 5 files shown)


src/_scss/pages/keyword/_searchBar.scss
```diff
@@ -8,54 +8,60 @@
...
                 }
             }
         }
     }
 
-    .info-text {
+    .keyword-search-bar__info {
         font-size: $small-font-size;
         line-height: rem(26);
         padding-left: rem(10);
```
kevinli-work: Should this padding be applied in mobile view? It offsets only the left side in when stacked top/bottom.

PR #648, opened 2018-02-07
Feature Account Landing

ebdabbs:
https://federal-spending-transparency.atlassian.net/browse/DEV-367

- Implements the Federal Account landing page at `/federal_account`
    - Search feature out of scope (hidden)
    - Defaults to sorting by descending budgetary resources
    - [Gist](https://gist.github.com/ebdabbs/bf7946a88ccd3b7a21e6c4a07bcc883b)
- Enables the Federal Accounts link in navigation
- Converts the following shared components to BEM classes:
    - Pagination (also affects spending explorer table view)
    - Landing page layout (also affects agency landing page)
    - Landing page search bar (also affects agency landing page)

- [x] [Endpoint](https://github.com/fedspendingtransparency/usaspending-api/pull/947) merged
- [x] Code Review
- [x] Design Review


src/_scss/components/_pagination.scss
```diff
@@ -1,6 +1,6 @@
 .pagination {
     margin: rem(15);
```
kevinli-work: You should not have margin declarations in your block level classes:
https://en.bem.info/methodology/quick-start/#block
> The block shouldn't influence its environment, meaning you shouldn't set the external geometry (margin) or positioning for the block.

An alternative would be to either wrap the children inside an element, which has margins, or to have the pagination component's parent component provide the margins.

src/_scss/components/_pagination.scss
```diff
@@ -1,6 +1,6 @@
...
-    .results-text {
+    .pagination__totals {
         @include display(flex);
         @include flex-direction(row);
         float: left;
```
kevinli-work: This `float: left` and the next `float: right` are a weird mix of flexbox and floating. Can you have the parent block be a flexbox, with `totals` and `pager` being flex items within that container that, in turn, act as flex containers for their children?

src/_scss/components/_pagination.scss
```diff
@@ -11,8 +11,8 @@
         @include flex-direction(row);
         float: right;
         margin: 0;
-        li {
-            button {
+        .pager__item {
```
kevinli-work: You need declare a block before declaring its member elements.

src/js/components/sharedComponents/Pagination.jsx
```diff
@@ -142,20 +156,26 @@ export default class Pagination extends React.Component {
...
         return (
             <div className="pagination">
-                <div className="results-text">
+                <div className="pagination__totals">
                     {resultsText}
                 </div>
                 <ul className="pager">
-                    <li className={pager.currentPage === 1 ? 'disabled' : ''}>
-                        <button onClick={() => this.setPage(pager.currentPage - 1)}>{`<`}</button>
+                    <li className={`pager__item ${pager.currentPage === 1 ? 'pager__item_disabled' : ''}`}>
```
kevinli-work: When buttons are disabled, can you set the `disabled` attribute on the button itself for accessibility? Also, the buttons still appear to be clickable at the moment, so you can go to negative pages.

src/_scss/components/_pagination.scss
```diff
@@ -24,15 +24,15 @@
...
+            &.pager__item_active {
+                .pager__button {
                     background-color: $color-primary-alt-dark;
                     color: $color-white;
                     border-color: $color-primary-alt-dark;
                 }
             }
-            &.disabled {
-                button {
+            &.pager__item_disabled {
```
kevinli-work: A more BEM-compliant approach would be to move the `disabled` modifier class to be a member of the `pager__button` element rather than its parent. This will involve moving the JSX code to set the disable modifier class on the `<button>` instead of the parent `<li>`.

kevinli-work: A similar, but opposite idea applies for setting the active page.

src/_scss/components/_pagination.scss
ebdabbs: `pager` is a block (buttons without the `# - ## of ## results` text)

src/_scss/components/_pagination.scss
ebdabbs: Moved `disabled` modifier class to the button

src/js/components/sharedComponents/Pagination.jsx
ebdabbs: Set a disabled attribute on the buttons

src/_scss/components/_pagination.scss
ebdabbs: Removed `float`s and added flexbox properties

src/_scss/components/_pagination.scss
ebdabbs: Removed positioning from the block; added margins in the spending explorer table file

src/js/containers/accountLanding/AccountLandingContainer.jsx
```diff
@@ -0,0 +1,208 @@
...
+            // Format budgetary resources
+            const formattedCurrency =
+                MoneyFormatter.formatMoneyWithPrecision(item.budgetary_resources, 0);
+
+            let formattedAgency = item.managing_agency;
+            if (item.managing_agency_acronym) {
+                formattedAgency = `${item.managing_agency} (${item.managing_agency_acronym})`;
+            }
+
+            const account = {
```
kevinli-work: Can you adapt this to follow our new `BaseXXX` object approach? You can take a look in `models/v2/BaseFederalAccountAwardRow.js` and `containers/account/awards/AccountAwardsContainer.jsx` for an example.

Basically, you would create a `BaseFederalAccountLandingRow` object with a `parse` function for ingesting the API response and handle null/empty values in that function. For derived fields (such as the combined agency name + agency abbrev col), you would use getter functions to concatenate (and determine if you can, based on if abbreviation is available) on the fly.

https://gist.github.com/kevinli-work/cfb001398a15db1afa8168fcbc4b0fb1

tests/containers/accountLanding/AccountLandingContainer-test.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+        expect(parseAccounts).toHaveBeenCalledTimes(1);
+    });
+
+    describe('showColumns', () => {
+        it('should build the table', async () => {
+            // mount the container
+            const container = mount(<AccountLandingContainer />);
+
+            await container.instance().accountsRequest.promise;
```
kevinli-work: Do you need to wait for the API request when testing the columns?
Also, could you do a shallow mount and explicitly call `showColumns` instead?

tests/containers/accountLanding/AccountLandingContainer-test.jsx
```diff
@@ -0,0 +1,117 @@
...
+                }
+            ];
+
+            expect(container.state().columns).toEqual(expectedState);
+        });
+    });
+
+    describe('parseAccounts', () => {
+        it('should parse the API response and update the container state', () => {
+            const container = mount(<AccountLandingContainer />);
```
kevinli-work: Where we are not testing the component lifecycle, we should use `shallow` instead in order to explicitly call (only) the method we are testing.

tests/containers/accountLanding/AccountLandingContainer-test.jsx
```diff
@@ -0,0 +1,117 @@
...
+    });
+
+    describe('updateSort', () => {
+        it('should update the state and make an API request', async () => {
+            // mount the container
+            const container = mount(<AccountLandingContainer />);
+            const parseAccounts = jest.fn();
+            container.instance().parseAccounts = parseAccounts;
+
+            await container.instance().accountsRequest.promise;
```
kevinli-work: In this and the next test, I don't believe we should be making API calls and parsing responses. These tests should be narrowly focused: given an explicitly provided input to the method being tested, what is the expected behavior/output?

For things like pagination, we can explicitly set the container's state/props via Enzyme's `setState` or `setProps` (http://airbnb.io/enzyme/docs/api/ShallowWrapper/setState.html), call the method being tested with a specific input, and then validate either the method's return value or the container's new state/props (http://airbnb.io/enzyme/docs/api/ShallowWrapper/state.html).

This ensures that our tests are specific (so when they fail, we know the exact behavior that is failing rather than "something is broken in the container"). The other behavior, like calling the API or parsing its response, should be covered by the tests for those methods and *this* test will rely on those tests to assume they are functioning correctly.

PR #660, opened 2018-02-08
Update - Advanced Search Hover

mab6bf:
- Updates text and link within the hover on the Advanced Search page Download button when it's in a disabled state
- Only shows the hover if the user has applied filters and there are more than 500,000 results

Issue: https://federal-spending-transparency.atlassian.net/browse/DEV-431

As there are fewer than 500,000 records within the Skinny Dev database, test the hover functionality by changing [this line](https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/containers/search/SearchContainer.jsx#L380) to `downloadAvailable: false`.

- [x] Code Review
- [x] Design Review


geoffgeiling:
Design Review approved

PR #666, opened 2018-02-16
Update download center urls

ebdabbs:
- Changes `bulk_download` and `bulk_download/award_data_archive` to `download_center/custom_award_data` and `download_center/award_data_archive`, respectively
- `/download_center` redirects to `download_center/custom_award_data`

### DIFF
--- a/src/js/containers/router/RouterRoutes.jsx
+++ b/src/js/containers/router/RouterRoutes.jsx
@@ -120,8 +120,9 @@ const routes = {
             }
         },
         {
-            path: '/bulk_download',
-            parent: '/bulk_download',
+            path: '/download_center',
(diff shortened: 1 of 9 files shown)


src/js/containers/router/RouterRoutes.jsx
```diff
@@ -120,17 +120,19 @@ const routes = {
             }
         },
         {
-            path: '/bulk_download',
-            parent: '/bulk_download',
+            path: '/download_center',
+            parent: '/download_center',
+            silentlyUpdate: true,
```
kevinli-work: Should this be `silentlyUpdate`? It seems to me that you would want the pages to reload completely when the URL changes between types.

PR #673, opened 2018-02-23
Feature Redirect Modal

ebdabbs:
- Adds a reusable redirect modal to the site
- Modal is triggered by the `Database Snapshot` links in navigation, the homepage download section, and the download center sidebar

- [x] Code Review
- [x] Design Review


kevinli-work:
The homepage link to the Database Snapshots is not handled by the modal.

src/_scss/pages/redirect/redirectModal.scss
```diff
@@ -0,0 +1,51 @@
+@import "all";
```
kevinli-work: Why is this not in the `modals` directory we have in `_scss/pages`?

src/_scss/pages/redirect/redirectModal.scss
```diff
@@ -0,0 +1,51 @@
...
+                width: rem(16);
+                fill: $color-white;
+            }
+        }
+    }
+    .redirect-modal__body {
+        text-align: center;
+        padding: rem(30) rem(45);
+        .redirect-modal__title {
+            @include display(inline-flex);
```
kevinli-work: What is the reasoning for using `inline-flex` instead of `flex`? It seems to me that you would want the title element to be block-positioned so the subsequent body content always appears below it.

kevinli-work: https://www.w3.org/TR/css-flexbox-1/#flex-containers

src/js/helpers/redirectHelper.js
```diff
@@ -0,0 +1,16 @@
...
+
+import storeSingleton from 'redux/storeSingleton';
+import * as redirectModalActions from 'redux/actions/redirectModal/redirectModalActions';
+
+
+export const showRedirectModal = (url) => {
+    storeSingleton.store.dispatch(redirectModalActions.showModal(url));
+};
+
+export const hideRedirectModal = () => {
```
kevinli-work: Do we actually need this? Isn't the only way to close the modal from inside the modal (which has access to `hideModal` via `react-redux`)?

src/js/components/sharedComponents/RedirectModal.jsx
```diff
@@ -0,0 +1,72 @@
...
+    url: PropTypes.string
+};
+
+export default class RedirectModal extends React.Component {
+    render() {
+        return (
+            <Modal
+                mounted={this.props.mounted}
+                onExit={this.props.hideModal}
+                titleText="You're Leaving a Bureau of the Fiscal Service website."
```
kevinli-work: Weird capitalization of the `L`

src/js/components/sharedComponents/header/DropdownItem.jsx
```diff
@@ -6,53 +6,85 @@
...
+                <a
+                    className="nav-children__link"
+                    href={this.props.url}
+                    target="_blank"
+                    rel="noopener noreferrer">
+                    {this.props.label}
+                </a>
+            );
+        }
+        else if (this.props.enabled && this.props.externalLink) {
```
kevinli-work: What is the use-case this is handling?

src/_scss/pages/redirect/redirectModal.scss
ebdabbs: Moved to the modals directory

src/js/helpers/redirectHelper.js
ebdabbs: Removed `hidRedirectModal`

src/js/components/sharedComponents/header/DropdownItem.jsx
ebdabbs: The Agency Submission Files link opens in a new tab but does not trigger the modal. This handles the Database Snapshots link, which does trigger the modal.

gitapablabah:
It's super close the main headline needs to be a little larger per the mock up attached. 
modal_mockup
TY!

gitapablabah:
The design edit to the headline looks great. Approved. Thank you!

PR #679, opened 2018-02-26
Fix - Header Mobile and Tablet Views

mab6bf:
- Fixes mobile and tablet views for the updated header bar in the Award and Federal Account Profile pages

Bug: https://federal-spending-transparency.atlassian.net/browse/DEV-530

- [x] Code Review
- [x] Design Review

### DIFF
--- a/src/_scss/pages/account/header/_options.scss
+++ b/src/_scss/pages/account/header/_options.scss
@@ -2,14 +2,18 @@
     @import "../../../mixins/moreOptionsBtn";
     @include display(flex);
     @include justify-content(space-between);
-    padding-top: $global-pad-mrg;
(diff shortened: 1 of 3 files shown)


src/_scss/pages/account/header/_options.scss
```diff
@@ -2,14 +2,18 @@
...
+
+    @include flex(1 1 auto);
+
     ul.summary-status-items {
         @include unstyled-list;
+        padding: rem(5) 0;
+
         li {
-            display: inline-block;
+            @include display(flex);
```
ebdabbs: Since we're using flex now, can we get rid of `float: left;` and `text-align: right;` in the child elements? Same goes for tablet view.

PR #681, opened 2018-02-26
Fix - Spending Explorer Table Key Names

mab6bf:
- Fixes Spending Explorer table unique key issue with duplicate names
- Fixes bug in showing cents in treemap tooltips for awards in the THOUSANDS range


src/js/helpers/moneyFormatter.js
```diff
@@ -156,18 +156,24 @@ export const calculateUnitForSingleValue = (value) => {
 export const formatTreemapValues = (value) => {
     // Format the ceiling and current values to be friendly strings
     const units = calculateUnitForSingleValue(value);
+    const isThousand = units.unit === unitValues.THOUSAND;
```
kevinli-work: What about less than a thousand?

mab6bf: Fixed

PR #729, opened 2018-03-13
Update award models

ebdabbs:
https://federal-spending-transparency.atlassian.net/browse/DEV-591

- Implements v2 award models
- Shows `--` instead of `Not Available` for null values
- Refactors Award Summary page components
- Adds tests for v2 models
- Removes old models


src/js/models/v2/awards/BaseAwardRecipient.js
```diff
@@ -0,0 +1,78 @@
...
+    populate(data) {
+        this.internalId = data.legal_entity_id || '';
+        this.name = data.recipient_name || 'Unknown';
+        this.duns = data.recipient_unique_id || '--';
+        this.parentDuns = data.parent_recipient_unique_id || '--';
+        this._businessTypes = data.business_types_description || '';
+        this._businessCategories = parseBusinessCategories(data);
+
+        // Recipient Location
+        let location = {};
```
kevinli-work: Should `location` default to `{}` or would it be better to use `null` so we can use truthy checks to see if it's available?
Alternatively, should it just populate a blank CoreLocation object if there's no location data provided?

src/js/dataMapping/contracts/additionalDetails.js
```diff
@@ -3,416 +3,230 @@
...
  */
 
-import * as MoneyFormatter from 'helpers/moneyFormatter';
-
 export const agencyFields = [
     {
         label: 'Awarding Agency',
-        field: 'awarding_agency_name'
+        field: '__special',
+        parse: (data) => data.awardingAgency.name
```
kevinli-work: When a `parse` function is no longer needed because there is no parsing logic, we should remove the function and update the `field` value to point to the model's property.

src/js/models/v2/awards/additionalDetails/BaseContractAdditionalDetails.js
```diff
@@ -0,0 +1,81 @@
...
+    populate(data) {
+        this.pricingCode = data.type_of_contract_pricing || '';
+        this.pricing = data.type_of_contract_pric_desc || '';
+
+        // Parent Award Details
+
+        this.idvType = data.idv_type || '--';
+        this.idcType = data.type_of_idc || '--';
+        this.idvAgencyId = data.referenced_idv_agency_iden || '--';
+        this.multipleIdv = data.multiple_or_single_award_i || '--';
```
kevinli-work: This looks like the old code points to `multiple_or_single_aw_desc` for this field.

src/js/models/v2/awards/additionalDetails/BaseContractAdditionalDetails.js
```diff
@@ -0,0 +1,81 @@
...
+
+const BaseContractAdditionalDetails = {
+    populate(data) {
+        this.pricingCode = data.type_of_contract_pricing || '';
+        this.pricing = data.type_of_contract_pric_desc || '';
+
+        // Parent Award Details
+
+        this.idvType = data.idv_type || '--';
+        this.idcType = data.type_of_idc || '--';
```
kevinli-work: This looks like it used be `type_of_idc_description`.

src/js/models/v2/awards/additionalDetails/BaseContractAdditionalDetails.js
```diff
@@ -0,0 +1,81 @@
+/**
+ * BaseContractAdditionalDetails.js
+ * Created by Lizzie Salita 3/5/18
+ */
+
+const BaseContractAdditionalDetails = {
```
kevinli-work: Many of these fields were remapped in the original `additionalDetails.js` file, so check that file to ensure you're looking at the right response fields.

src/js/models/v2/awards/additionalDetails/BaseContractAdditionalDetails.js
```diff
@@ -0,0 +1,81 @@
...
+    populate(data) {
+        this.pricingCode = data.type_of_contract_pricing || '';
+        this.pricing = data.type_of_contract_pric_desc || '';
+
+        // Parent Award Details
+
+        this.idvType = data.idv_type || '--';
```
kevinli-work: `idv_type_description`

src/js/models/v2/awards/additionalDetails/BaseContractAdditionalDetails.js
```diff
@@ -0,0 +1,81 @@
...
+        this.idcType = data.type_of_idc || '--';
+        this.idvAgencyId = data.referenced_idv_agency_iden || '--';
+        this.multipleIdv = data.multiple_or_single_award_i || '--';
+
+        // Competition Details
+        this.solicitationId = data.solicitation_identifier || '--';
+        this.solicitationProcedures = data.solicitation_procedures || '--';
```
kevinli-work: This should use `solicitation_procedur_desc` if truthy, otherwise fall back to `solicitation_procedures` before using the `--`.

kevinli-work: Please review the rest of this file to ensure the correct mappings.

src/js/models/v2/awards/BaseFinancialAssistance.js
```diff
@@ -0,0 +1,117 @@
...
+    };
+    const placeOfPerformance = Object.create(CoreLocation);
+    placeOfPerformance.populateCore(placeOfPerformanceData);
+    this.placeOfPerformance = placeOfPerformance;
+
+    if (data.awarding_agency) {
+        const awardingAgencyData = {
+            name: data.awarding_agency.toptier_agency && data.awarding_agency.toptier_agency.name,
+            subtierName: data.awarding_agency.subtier_agency && data.awarding_agency.subtier_agency.name,
+            officeName: data.latest_transaction && data.latest_transaction.assistance_data
```
kevinli-work: Should office name be `awarding_office_name` instead of code? (etc for funding)

src/js/models/v2/awards/BaseContract.js
```diff
@@ -0,0 +1,125 @@
...
+        placeOfPerformance.populateCore(placeOfPerformanceData);
+        this.placeOfPerformance = placeOfPerformance;
+    }
+
+    if (data.awarding_agency) {
+        const awardingAgencyData = {
+            name: data.awarding_agency.toptier_agency && data.awarding_agency.toptier_agency.name,
+            subtierName: data.awarding_agency.subtier_agency && data.awarding_agency.subtier_agency.name,
+            officeName: (data.latest_transaction && data.latest_transaction.contract_data)
+                && (data.latest_transaction.contract_data.awarding_office_name || data.latest_transaction.contract_data.awarding_office_code)
```
kevinli-work: Office code or office name? Also applies to funding agency

src/js/dataMapping/contracts/additionalDetails.js
ebdabbs: I'm using parse functions for nested fields when the data are on different levels or in different objects (in this case `awardingAgency` and `fundingAgency`). Let me know if there is a better way to access/ organize them.

src/js/models/v2/CoreLocation.js
```diff
@@ -64,7 +64,8 @@ const CoreLocation = {
     get fullAddress() {
         const line1 = (this.streetAddress && `${this.streetAddress}`) || '';
         const line2 = (this.regionalAddress && `${this.regionalAddress}\n`) || '';
-        return `${line1}${line2}${this.congressionalDistrict}` || '--';
+        const line3 = (this.congressionalDistrict && `Congressional District: ${this.congressionalDistrict}`) || '';
+        return `${line1}${line2}${line3}` || '--';
```
kevinli-work: This logic is fine, but it doesn't look like the `\n` are rendering on the actual page. I think we need to add `white-space: pre-wrap;` in the CSS to the relevant items.
https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

src/js/models/v2/awards/BaseAwardRecipient.js
```diff
@@ -18,10 +18,17 @@ const parseBusinessCategories = (data) => (
...
-        const amount = formatMoney(data[`officer_${i}_amount`]) || 0;
-        executiveCompensation[`officer${i}`] = `${name} - ${amount}`;
+    if (data) {
+        for (let i = 1; i < 6; i++) {
+            const name = data[`officer_${i}_name`] || '';
+            const amount = formatMoney(data[`officer_${i}_amount`]) || 0;
+            executiveCompensation[`officer${i}`] = `${name} - ${amount}`;
+        }
+    }
+    else {
```
kevinli-work: I don't think the properly handles the case where some executive compensation is reported, but not all 6 (ie, 4 are reported, the last two slots should be `--`).

src/js/models/v2/awards/BaseContract.js
```diff
@@ -81,15 +81,15 @@ BaseContract.populate = function populate(data) {
         this.additionalDetails = additionalDetails;
     }
 
-    this.parentAward = data.parent_award_piid || '--';
+    this.parentAward = data.parent_award_id || '--';
```
kevinli-work: We need to look at **both** the `additionalDetails.js` mapping and the original `AwardSummary.js` model object. In the original logic, the container created an `AwardSummary` object from the API response, which remapped some API fields to custom properties. Where fields were not remapped, the original API field would be appended to the object. The presentation component would then **re-remap** those fields again via the `additionalDetails.js` file. So when we are updating the code to be simpler (and have only one remapping), we need to consult `additionalDetails` first, then trace back to `AwardSummary` to see if it is a double-remapped field or if it is an original API field.

In this case, the `parent_award_id` field in `additionalDetails.js` is pointing to a field that was remapped by `AwardSummary` from the API's `parent_award_piid` field.

From:
https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/models/results/award/AwardSummary.js#L223
to:
https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/models/results/award/AwardSummary.js#L478

src/js/components/award/DetailRow.jsx
```diff
@@ -53,24 +42,23 @@ export default class DetailRow extends React.Component {
     }
 
     render() {
-        let value = null;
-        if (this.props.title === "Primary Place of Performance") {
-            value = this.props.value.split('\n').map((item, key) =>
-                <span key={key}>{item}<br /></span>
-            );
+        const maxChars = SummaryPageHelper.maxDescriptionCharacters;
```
kevinli-work: Not all fields are expandable; only the award and CFDA descriptions. The overflow/expandable feature should probably only be enabled based on a prop.

src/js/components/award/contract/ContractDetails.jsx
```diff
@@ -4,185 +4,26 @@
...
-
-        // char count
-        let seeMore = false;
-        if (award.description.length > this.props.maxChars) {
-            seeMore = true;
+    render() {
+        const award = this.props.selectedAward;
+        let awardType = award.awardType;
+        if (award.category === 'idv') {
+            awardType = idvAwardTypes[this.props.selectedAward.awardType];
```
kevinli-work: This logic is broken for IDVs because `awardType` was remapped to `latest_transaction.contract_data.idv_type_description` instead of `latest_transaction.contract_data. idv_type`

kevinli-work: More generally, because `DetailRow` performs a length check on the string, we must ensure that whatever we pass into `DetailRow` has a length value (not `null` or `undefined`) or the page will fail to render.

src/js/components/award/RecipientAddress.jsx
```diff
@@ -7,68 +7,25 @@ import React from 'react';
...
-
+        const location = this.props.location;
         return (
             <li className={this.props.type}>
                 <div className="item-label">
                     Address
                 </div>
                 <div className="item-value">
-                    {street}
+                    {location.streetAddress}
```
kevinli-work: Any reason this wasn't updated to use CoreLocation's `fullAddress` like it is in `ContractDetails`? I think the logic here should match ContractDetails for consistency (or vice versa).

src/js/components/award/RecipientAddress.jsx
```diff
@@ -7,68 +7,25 @@ import React from 'react';
...
                 </div>
                 <div className="item-value">
-                    {cityState} {country} {recipient.recipient_zip_postal}
+                    {location.regionalAddress}
                 </div>
-                {district}
+                {location.congressionalDistrict}
```
kevinli-work: If we keep this approach, we need to label this `Congressional District: {location.congressionalDistrict}` for consistency with the current site and also ContractDetails.

src/js/helpers/timeRangeHelper.js
```diff
@@ -0,0 +1,44 @@
...
+
+export const convertDatesToRange = (startDate, endDate) => {
+    if ((startDate && endDate) && (moment.isMoment(startDate) && moment.isMoment(endDate))) {
+        const duration = moment.duration(endDate.diff(startDate));
+        const years = duration.years();
+        const months = duration.months();
+        let yearString = '';
+        let monthString = '';
+
+        if (!isNaN(years)) {
```
kevinli-work: What is the condition `!isNaN(years)` is trying to catch? If it is to detect <1 year, wouldn't `years` be 0?

src/js/helpers/timeRangeHelper.js
```diff
@@ -0,0 +1,44 @@
...
+                }
+            }
+
+            if (monthString && yearString) {
+                return `(${yearString}, ${monthString})`;
+            }
+
+            return `(${monthString}${yearString})`;
+        }
+        return '';
```
kevinli-work: This doesn't seem to catch durations of less than 1 month. Currently durations of less than 1 month return empty parens `()`.

kevinli-work: Look at `#/award/612`

src/js/components/award/contract/ContractDetails.jsx
kevinli-work: Look at `#/award/637`

src/js/models/v2/awards/BaseFinancialAssistance.js
ebdabbs: Changed to `awarding_office_name` and `funding_office_name`

src/js/components/award/DetailRow.jsx
ebdabbs: Added an `overflow` prop for award & CFDA descriptions

kevinli-work:
I was wrong about subsidy cost, it should be reverted back to the top-level `total_subsidy_cost` as you originally coded.

tests/models/CoreLocation-test.js
```diff
@@ -0,0 +1,63 @@
...
+   });
+   it('should use province as the state/province when state is not available', () => {
+       expect(foreignLocation.stateProvince).toEqual('Quebec');
+   });
+   it('should use the state code as state/province when city is available', () => {
+       expect(location.stateProvince).toEqual('IN');
+   });
+   it('should format the full address', () => {
+       expect(location.fullAddress).toEqual('602 Trumball Street\nApt 2\nPawnee, IN 12345\nCongressional District: IN-04');
+   });
```
kevinli-work: Can you add some more cases to test address formatting wen there are missing fields? This will help us be better about catching bugs in the future.

tests/models/awards/BaseContract-test.js
```diff
@@ -0,0 +1,146 @@
...
+           expect(contract.fundingAgency).toBeFalsy();
+       });
+       it('should format toptier and subtier names', () => {
+          expect(contract.awardingAgency.name).toEqual('Department of Sandwiches');
+          expect(contract.awardingAgency.subtierName).toEqual('Department of Subs');
+          expect(contract.awardingAgency.officeName).toEqual('Office of Cheesesteak');
+       });
+    });
+    describe('Place of Performance', () => {
+        it('should format the regional address', () => {
```
kevinli-work: Most of this behavior should be covered by `CoreLocation`'s tests. Instead we should validate that the `placeOfPerformance` property is an object with `CoreLocation` in its prototype chain. This can be done using [Object.getPrototypeOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getPrototypeOf)

src/js/models/v2/BaseFederalAccountAwardRow.js
```diff
@@ -19,12 +19,10 @@ const BaseFederalAccountAwardRow = {
         this.awardingToptierAgency = data.awarding_agency.toptier_agency.name || '';
         this.awardingSubtierAgency = data.awarding_agency.subtier_agency.name || '';
         this._issuedDate = parseDate(data.date_signed) || '';
-        this._loanValue = (data.latest_transaction
-            && data.latest_transaction.assistance_data
-            && parseFloat(data.latest_transaction.assistance_data.face_value_loan_guarantee)) || 0;
         this._subsidyCost = (data.latest_transaction
```
kevinli-work: I believe this also needs to be updated to `total_subsidy_cost`

tests/models/awards/BaseContract-test.js
```diff
@@ -77,70 +78,71 @@ describe('BaseContract', () => {
...
     describe('Place of Performance', () => {
-        it('should format the regional address', () => {
-            expect(contract.placeOfPerformance.regionalAddress).toEqual('Pawnee, IN 12345');
-        });
-        it('should format the full address', () => {
-           expect(contract.placeOfPerformance.fullAddress).toEqual('Pawnee, IN 12345\nCongressional District: IN-04')
+        it('should be an object with CoreLocation in its prototype chain', () => {
+            const locationObject = Object.create(CoreLocation);
+
+            expect(Object.getPrototypeOf(locationObject)).toEqual(Object.getPrototypeOf(contract.placeOfPerformance));
```
kevinli-work: A clearer way of testing this would be
```
expect(Object.getPrototypeOf(contract.placeOfPerformance)).toEqual(CoreLocation);
```

kevinli-work: Can we add prototype tests to the other subobjects (agency/recipient/etc) as well?

PR #734, opened 2018-03-16
Fix: Pass all the children of a parent recipient type when it is selected

kevinli-work:
https://federal-spending-transparency.atlassian.net/browse/DEV-624

* Sends all the child recipient type values to the API when a parent recipient type filter is selected on Advanced Search
* When a parent recipient type is selected, any previously selected child recipient type is overwritten with the parent recipient type to prevent duplicates in the API call and the top filter bar


kevinli-work:
Retracted.

PR #749, opened 2018-03-28
Download recipient location filter

ebdabbs:
https://federal-spending-transparency.atlassian.net/browse/DEV-641

- Implements an optional recipient location filter for Custom Award Data download
    - User can filter by United States or All Foreign Countries
    - When United States is selected, user can filter by state
- Implements a tooltip for the disabled download button
- Converts download filters to BEM classes

- [x] Code review
- [x] Design review

### DIFF
--- a/src/js/components/bulkDownload/awards/filters/LocationFilter.jsx
+++ b/src/js/components/bulkDownload/awards/filters/LocationFilter.jsx
@@ -6,7 +6,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import * as Icons from 'components/sharedComponents/icons/Icons';
+import { CheckCircle } from 'components/sharedComponents/icons/Icons';
+import EntityDropdown from 'components/search/filters/location/EntityDropdown';
 
 const propTypes = {
     states: PropTypes.array,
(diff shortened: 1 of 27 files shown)


gitapablabah:
Looks great! Minor tweaks.
1) Can you spell out the States similar to Advanced Search
2) Can you replace the 3 dashes in the second dropdown to 'None'

Thank you!

gitapablabah:
Correction on #2: Can you replace the 3 dashes in the second dropdown to 'All'

src/js/redux/reducers/bulkDownload/bulkDownloadReducer.js
```diff
@@ -25,6 +25,13 @@ export const initialState = {
             id: '',
             name: 'Select a Sub-Agency'
         },
+        location: {
+            country: {
+                code: '',
+                name: ''
+            },
+            state: ''
```
kevinli-work: This should probably also be an object, in order to retain both the full name and state code.

src/js/components/bulkDownload/awards/filters/LocationFilter.jsx
```diff
@@ -0,0 +1,219 @@
...
+        const countries = countryOptions.map((country) => (
+            <li
+                className="field-item indent"
+                key={`field-${country.code}`}>
+                <button
+                    className="item-button"
+                    title={country.name}
+                    aria-label={country.name}
+                    value={country.code}
+                    name={country.name}
```
kevinli-work: You should not be saving the country name in the `name` HTML attribute nor receiving it via the event `target.name` value. This is because `name` is intended to label the control's field when a form is sent to a server, not for holding a value. See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes

kevinli-work: A more appropriate approach would be to store the code in the `value` attribute, and then look up the code to retrieve the associated data in your event handler.

gitapablabah:
Looks really good, thank you!

PR #758, opened 2018-04-12
Fix Custom Award Data Date Range

dtrinh50:
https://federal-spending-transparency.atlassian.net/browse/DEV-802

- Removed _today_ from the list of predefined data ranges on the Custom Award Data page.


gitapablabah:
Add 'last 60 days' in the place where the gap is. See mock up. Thank you!

bfs_cad_revision_041618

PR #761, opened 2018-04-24
Feature Three Pane Layout

dtrinh50:
https://federal-spending-transparency.atlassian.net/browse/DEV-826

- Implemented 3 pane layout for featured content.

- [ ] Code Review
- [x] Design Review

### DIFF
--- a/src/_scss/pages/homepage/features/_featurePane.scss
+++ b/src/_scss/pages/homepage/features/_featurePane.scss
@@ -1,0 +1,88 @@
+.feature-pane {
+    .feature-pane__wrapper {
+        display: block; 
+        margin: rem(60) 0 rem(100);
+        width: 100%;
+
+        @include media($medium-screen) {
+            margin: rem(120) 0 rem(60);
+        }
+    }
+
+    .feature-pane__title {
+        width: 100%; 
+        text-align: center; 
+        border-bottom: 1px solid rgb(214,215,217);
+        line-height: rem(0.1);
+        font-size: rem(16);
+        margin: 10px 0  20px;
+        color: #717171;
+
+        & span {
+            background:#fff; 
+            padding:0 10px; 
+        }
+    }
+
+    .feature-pane__content-wrapper {
+        display: flex;
+        justify-content: space-around;
+        margin-bottom: rem(30);
+        flex-wrap: wrap;
+    }
+
+    .feature-pane__content {
+        flex-basis: 100%;
+        text-align: center;
+        font-size: rem(14);
+        line-height: rem(17);
+
+        @include media($medium-screen) {
+            flex-basis: 32%;
+            max-width: 32%;
+        }
+    }
+    
+    .feature-pane__content-divider {
+        display: none;
+        @include media($medium-screen) {
+            display: block;
+            border: 1px solid rgba(214,215,217,0.4);
+            height: 180px;
+            align-self: center;
+        }
+        
+    }
+
+    .feature-pane__content-title {
+        font-size: rem(24);
+        font-weight: bold;
+        text-align: center;
+        color: #5B616B;
+    } 
+
+    .feature-pane__button-wrapper {
+        display: flex;
+        justify-content: center;
+    }
+    
+    .feature-pane__bottom-divider {
+        border: 1px solid rgb(214,215,217);
+    }
+
+    .feature-pane_mobile_content_divider {
+        display: block;
+        width: 100%;
+        margin-top: rem(30);
+        @include media($medium-screen) {
+            display: none;
+        }
+    }
+
+    .feature-pane__button {
+        color: #0071BC;
+        background-color: transparent;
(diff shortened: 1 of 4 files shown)


gitapablabah:
Looks really close to done. The only issue I see is that the bottom horizon rule is thicker than the top one. I checked in Chrome, Firefox & Safari.

src/js/components/homepage/features/PaneFeature.jsx
```diff
@@ -0,0 +1,50 @@
...
+ * PaneFeature.jsx
+ * Created by David Trinh 4/19/18
+ */
+
+import React from 'react';
+
+const PaneFeature = () => (
+    <div className="feature-pane">
+        <div className="feature-pane__wrapper">
+            <h2 className="feature-pane__title"><span>FEATURED CONTENT</span></h2>
```
kevinli-work: The title should be the content of the `h2` tag. Instead of styling the `h2` tag to be the grey border line and then styling the inner text using `span`, I would recommend making the text an `h2` (no span) and then wrapping the `h2` inside a `div` that provides the border styling.

Additionally, please write the title in normal sentence case ("Featured Content") and use `text-transform` to display it as upper case. This allows users who have trouble reading all caps text to override with a custom local stylesheet.

src/js/components/homepage/features/PaneFeature.jsx
```diff
@@ -0,0 +1,50 @@
...
+            <div className="feature-pane__content-wrapper">
+                <div className="feature-pane__content">
+                    <h2 className="feature-pane__content-title">Discover the Data Lab</h2>
```
kevinli-work: Since these titles are children of Feature Content, they should be `h3` tags to indicate that hierarchy.

src/js/components/homepage/features/PaneFeature.jsx
```diff
@@ -0,0 +1,50 @@
...
+                    <div className="feature-pane_content-text">
+                        <p>Data visualizations to help you understand government spending. See what our data can do.</p>
```
kevinli-work: Any reason that the `<p>` tags are wrapped inside a `<div>` tag rather than applying the class directly to the `<p>`?

src/_scss/pages/homepage/features/_featurePane.scss
```diff
@@ -0,0 +1,116 @@
...
+    .feature-pane__button-wrapper {
+        display: flex;
+        justify-content: center;
+    }
+    
+    .feature-pane__bottom-divider {
+        border: 1px solid rgba(214,215,217,0.4);
+    }
+
+    .feature-pane_mobile_content_divider {
```
kevinli-work: Elements should be separated from the block name with a double underscore `__`

src/_scss/pages/homepage/features/_featurePane.scss
```diff
@@ -0,0 +1,116 @@
...
+        color: #717171;
+
+        & span {
+            background:#fff; 
+            padding:0 10px; 
+        }
+    }
+
+    .feature-pane__content-wrapper {
+        display: flex;
```
kevinli-work: Our Sass code uses mixins for flexbox layout that adds vendor prefixes for older browsers. Please update to use these; for the most part, they are `@import [property]([value])`, so for example `@include display(flex)` or `@include justify-content(space-around)`

src/_scss/pages/homepage/features/_featurePane.scss
```diff
@@ -0,0 +1,116 @@
...
+        justify-content: space-around;
+        margin-bottom: rem(20);
+        flex-wrap: wrap;
+    }
+
+    .feature-pane__content {
+        flex-basis: 100%;
```
kevinli-work: Rather than use `flex-basis`, please use the `flex` [shorthand property](https://developer.mozilla.org/en-US/docs/Web/CSS/flex) instead. This makes grow and shrink factors explicit (IE11 may not use the defaults you're expecting with other browsers).

src/js/components/homepage/features/PaneFeature.jsx
dtrinh50: How about using the pseudo-elements after and before, instead of div?

src/js/components/homepage/features/PaneFeature.jsx
dtrinh50: I did it because I wanted to apply a margin bottom to the text, but I don't think margin bottom works with inline elements. I can just add margin top to the button wrapper instead.

dtrinh50:
@kevinli-work I made the requested updates. Please let me know if you have any other changes that is needed.

@AGuyNamedMarco The recent code change might have made slight unintentionally changes to the layout, please let me know if I need to fix anything.

gitapablabah:
There are no design issues with the latest code changes. It's good to go. TY!

PR #774, opened 2018-05-10
Feature Delta Text

dtrinh50:
https://federal-spending-transparency.atlassian.net/browse/DEV-971

- Updated the body copy

- Wrote code to handle null fiscal years

- Made additional UI changes to properly align filters on the page

- [ ] Design Review

- [ ] Code Review


src/_scss/pages/bulkDownload/archive/_archiveForm.scss
```diff
@@ -1,6 +1,12 @@
 .award-data-archive-form {
     background-color: #F3FAFC;
     padding: rem(15) rem(5);
+
+    .form-title__wrapper {
```
ebdabbs: Is this wrapper/ the flex properties necessary? We could add `padding-bottom: rem(10);` to the `.form-title` element for spacing.

src/_scss/pages/bulkDownload/archive/_archiveForm.scss
```diff
@@ -55,14 +69,14 @@
                 }
             }
         }
-        .submit-button {
-            padding: 0 rem(15);
+        .form__button {
+            padding-left: rem(15);
```
ebdabbs: I think we want to leave 15px padding on the right so that this button is the same distance from the edge of the blue box as the Agency select.

gitapablabah:
The UI looks good. TY! Doing this on my phone. Not able to check the Design Review box, so I emoji’d a thumbs up.

PR #786, opened 2018-05-22
State Award Breakdown Contract

dpb-bah:


api_contracts/contracts/state/StateProfile.md
```diff
@@ -41,3 +56,8 @@ This endpoint returns a high-level overview of a specific state or territory, gi
 + total_prime_amount: 300200000000 (required, number)
 + total_prime_awards: 327721 (required, number)
 + award_amount_per_capita: 916023.08 (required, number)
+
+## StateBreakdown(object)
```
ebdabbs: I would suggest using the line above "Award types include `contracts`, `grants`, `direct_payments`, `loans`, `other_financial_assistance`" here as a description for `type` and adding descriptions for `amount` (The aggregate value of awards of this type.) and `count` (The number of awards of this type.) for clearer documentation.

dpb-bah: Good call. Updated.

PR #788, opened 2018-05-22
Feature State Profile Award Breakdown

ebdabbs:
[DEV-941](https://federal-spending-transparency.atlassian.net/browse/DEV-941)

-  Implements the award breakdown section (treemap + table) of the state profile page
- Updates the state profile layout to use Flexbox instead of floats
- Includes a fix for `<hr>` divider elements showing up as three lines instead of one in browsers other than Chrome

- [ ] Code review
- [ ] Design review
- [ ] API merged


src/js/components/state/visualizations/awardBreakdown/AwardBreakdownTreeMap.jsx
```diff
@@ -0,0 +1,265 @@
...
+                windowWidth,
+                visualizationWidth: this.sectionWrapper.offsetWidth
+            });
+            if (this.props.awardBreakdown.children.length > 0) {
+                this.buildTree(this.props);
+            }
+        }
+    }
+
+    buildTree(treeProps) {
```
kevinli-work: Can you update this to follow the `buildVirtualChart` approach we use in the spending explorer? https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/components/explorer/detail/visualization/treemap/ExplorerTreemap.jsx#L54

Basically, instead of generating SVG components and saving them in state, we create JS objects that have all the calculated values the SVG components will need and then put the actual SVG in either standalone components or in the render output. The `buildVirtualChart` method would be limited in scope to doing just calculations rather than render related logic (but it can do color scale calculation logic).

src/js/containers/state/visualizations/awardBreakdown/AwardBreakdownContainer.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+        this.searchRequest.promise
+            .then((res) => {
+                this.searchRequest = null;
+
+                this.setState({
+                    inFlight: false
+                });
+
+                // Sum all amounts in the returned award types
```
kevinli-work: Could this be done in a parser function? This will also make it easier to test.

geoffgeiling:
Design Review Approved

PR #789, opened 2018-05-22
State Profile: Top Five

kevinli-work:
https://federal-spending-transparency.atlassian.net/browse/DEV-942

* Requires API endpoint: api/v2/spending_by_category/
* Adds Top Five tables to the state profile page

- [x] API endpoint merged: fedspendingtransparency/usaspending-api#1268
- [x] Code review
- [x] Design review


gitapablabah:
Feedback provided via PDF and posted to Slack #ux-frontend channel. TY.

src/js/components/state/topFive/TopFiveRow.jsx
```diff
@@ -15,20 +17,24 @@ const TopFiveRow = (props) => {
     const percentValue = (props.data._amount / props.total) * 100;
 
     const percent = isNaN(percentValue) ? '--' : `${Math.round(percentValue * 100) / 100}%`;
+    const formattedAmount = MoneyFormatter.formatStateValues(props.data._amount);
```
ebdabbs: See later comment about formatting logic. This should be done in the model so that `{props.data.amount}` is already formatted.

src/js/helpers/moneyFormatter.js
```diff
@@ -186,6 +186,39 @@ export const formatTreemapValues = (value) => {
     return `${formattedCurrency}${longLabel}`;
 };
 
+export const formatStateValues = (value) => {
```
ebdabbs: Can we move this logic to the model? moneyFormatter should be more generic helpers. Since this is specifically for Top 5 data, it can go in the getter function for `amount` in BaseStateCategoryResult.js. Or, if we format money in the same way (conditionally, based on amount) throughout the site, we can give this a more generic name for re-use.

gitapablabah:
Current design work approved for this sprint. We will need some design tweaks/hot fixes to the truncated data as soon as possible.

PR #790, opened 2018-05-23
Fixing Tabs in State Profile Contract

dpb-bah:


kevinli-work:
In the future, prefix the PR with "API Contract" so we can differentiate from frontend code changes.

PR #801, opened 2018-06-05
Enable Spending by Category

ebdabbs:
[DEV-832](https://federal-spending-transparency.atlassian.net/browse/DEV-832)
- Re-enables the Spending by Category visualization on the Advanced Search page 
    - Displays ten results instead of five
    - Removes Spending by Funding Agency option
- Refactored to comply with the current Redux structure & API format
    - Updates the API request to reflect the prime/sub-award toggle
    - The spending by category endpoint no longer takes a `category` param, which eliminates the need for separate containers for each "Spending by" option
- Uses a new v2 model to format the returned results 
- Updates copy per recommendations from the UX team

- [x] Code review
- [x] Design review


src/_scss/pages/search/results/visualizations/rank/_topSection.scss
```diff
@@ -1,18 +1,45 @@
 .visualization-top {
-    @include row();
-    margin-bottom: rem(20);
-    @import "components/visualizations/_description";
-    @import "components/visualizations/_period";
-
-    .no-scope {
-        width: 100%;
+    padding-bottom: rem(20);
+    clear: both;
```
dtrinh50: Are we clearing some float element here?

gitapablabah:
Set the max width of the yellow warning message to 720px. TY!

src/_scss/pages/search/results/visualizations/rank/_topSection.scss
ebdabbs: Yes, the `<hr>` element had a float-left property which I've removed.

AGuyNamedMarco:
Design review completed and approved.

I did notice the following items were not included but wanted to note them as future tasks (since they aren't trivial):

- In "Categories" tab's bar chart, reduce height of bar chart items
- In "Categories" tab's bar chart, Automatically reduce total bar chart height when there are less than 10 items returned
- Update pagination to use latest page numbered version

PR #806, opened 2018-06-12
Feature Add Outline to State Profile Flags

dtrinh50:
https://federal-spending-transparency.atlassian.net/browse/DEV-1131

- Added outline to state profile flags

- [x] Code Review

- [x] Design Review

### DIFF
--- a/src/_scss/pages/state/overview/_overview.scss
+++ b/src/_scss/pages/state/overview/_overview.scss
@@ -10,6 +10,10 @@
             width: rem(64);
             height: auto;
             margin-right: rem(28);
+
+            & img {
+                border: 1px solid #CCCCCC;
+            }
         }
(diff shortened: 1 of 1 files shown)


src/_scss/pages/state/overview/_overview.scss
```diff
@@ -10,6 +10,10 @@
             width: rem(64);
             height: auto;
             margin-right: rem(28);
+
+            & img {
+                border: 1px solid #CCCCCC;
```
ebdabbs: `img` doesn't need the ampersand in front here, since it is below `state-overview__flag` in the DOM tree.
Can you use the variable we have for `$color-gray-lighter`?

gitapablabah:
1px rule around flag looks good. Increased flag size a little.

PR #820, opened 2018-06-26
Close advanced search when complete

kojodoesdesign:
https://federal-spending-transparency.atlassian.net/browse/DEV-1139
- Modal is now properly hidden after downloading results from advanced search
- Modal also resets when reopened to allow the user to request a new download

### DIFF
--- a/src/js/components/search/modals/fullDownload/FullDownloadModal.jsx
+++ b/src/js/components/search/modals/fullDownload/FullDownloadModal.jsx
@@ -39,8 +39,8 @@ export default class FullDownloadModal extends React.Component {
         this.hideModal = this.hideModal.bind(this);
     }
 
-    componentDidUpdate(nextProps) {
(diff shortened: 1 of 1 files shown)


src/js/components/search/modals/fullDownload/FullDownloadModal.jsx
```diff
@@ -39,18 +39,19 @@ export default class FullDownloadModal extends React.Component {
         this.hideModal = this.hideModal.bind(this);
     }
 
-    componentWillReceiveProps(nextProps) {
+    componentDidUpdate(nextProps) {
```
ebdabbs: Thanks for updating this method! Could you use `prevProps` for the comparison instead of `nextProps`?

kojodoesdesign: fixed and updated branch!

PR #831, opened 2018-07-06
Added State Abbreviations for each state in State Profile

kojodoesdesign:
- States now display their abbreviations in the table
- Fixed a search issue which sometimes caused multiple states to be displayed when the user searched by a state abbreviation. 
The search error typically occured with these abbreviations:
ME
MA
MI
MO 
NE

- [x] Code Review
- [x] Design Review

### DIFF
--- a/src/js/components/stateLanding/table/StateLandingTable.jsx
+++ b/src/js/components/stateLanding/table/StateLandingTable.jsx
@@ -23,13 +23,19 @@ const propTypes = {
 const StateLandingTable = (props) => {
     const hideBody = props.loading || props.error || props.results.length === 0 ? 'state-list__body_hide' : '';
 
-    const body = props.results.map((row) => (
+    // Check if the search string is a known state abbreviation
+    const abbreviation = [props.results.find((obj) => obj.code === props.searchString)];
+    const abbrvResults = abbreviation[0] ? abbreviation : props.results;
+
+    const body = abbrvResults.map((row) => (
         <tr
             key={row.fips}
             className="state-list__body-row">
             <StateLinkCell
                 fips={row.fips}
                 name={row.name}
+                code={row.code}
+                isAbbrev={abbreviation}
                 searchString={props.searchString} />
             <td className="state-list__body-cell state-list__body-cell_right">
                 {row.amount}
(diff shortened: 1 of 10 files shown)


src/js/components/stateLanding/table/StateLinkCell.jsx
```diff
@@ -9,15 +9,29 @@ import reactStringReplace from 'react-string-replace';
...
-    searchString: PropTypes.string
+    searchString: PropTypes.string,
+    isAbbrev: PropTypes.array
 };
 
 export default class StateLinkCell extends React.Component {
     render() {
         let name = this.props.name;
+        const abbreviation = this.props.isAbbrev[0];
+        const stateWithAbbreviation = `${name}, ${this.props.code}`;
```
ebdabbs: I would do this formatting in the model (`models/v2/state/BaseStateLandingItem.js`) so that `this.props.name` already includes the abbreviation. Also please update to use parens instead of a comma.

src/js/components/stateLanding/table/StateLandingTable.jsx
```diff
@@ -23,13 +23,19 @@ const propTypes = {
 const StateLandingTable = (props) => {
     const hideBody = props.loading || props.error || props.results.length === 0 ? 'state-list__body_hide' : '';
 
-    const body = props.results.map((row) => (
+    // Check if the search string is a known state abbreviation
```
ebdabbs: We are already searching on the `code` (abbreviation) here: https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/containers/stateLanding/StateLandingContainer.jsx#L108, so when the code is added to `name` in the model (see comment below) it should automatically be highlighted.

AGuyNamedMarco:
Design review complete – passed.

Thanks, @kojodoesdesign!

src/js/models/v2/state/BaseStateLandingItem.js
```diff
@@ -7,12 +7,15 @@ import * as MoneyFormatter from 'helpers/moneyFormatter';
...
     populate(data) {
-        this.name = data.name || '';
+        this._name = data.name || '';
         this.fips = data.fips || '';
         this.code = data.code || '';
         this.type = data.type || '';
         this._amount = data.amount || 0;
     },
+    get name() {
+        return this.code ? `${this._name} (${this.code})` : this._name;
```
ebdabbs: Good use of `?` operator 👍

kojodoesdesign: Thanks!

PR #842, opened 2018-07-16
Recipient profile top 5 (front end)

kojodoesdesign:
https://federal-spending-transparency.atlassian.net/browse/DEV-1234

- Added Top 5 Tables to Recipient Profile page
- Upgraded Code to es6 class implementation
- Text is truncated in the same way as state profile pages so that each table is a consistent height
- Added loading, error, and "No data available" states to tables


kojodoesdesign:
Closed due to branch naming bug

PR #870, opened 2018-08-09
Update hide trendline

ebdabbs:
https://federal-spending-transparency.atlassian.net/browse/DEV-1399

- Hides the trendline + its axis & label by re-using state profile's time visualization component
    - Keeps the x-axis labels static on hover per Recipient Profile design review
    - Adds a prop for the color of the bars (can be removed later)


src/js/components/state/spendingovertime/StateTimeVisualization.jsx
```diff
@@ -67,7 +68,7 @@ export default class StateTimeVisualization extends React.Component {
 
         const legend = [
             {
-                color: '#708893',
+                color: this.props.color,
```
kojodoesdesign: I think having this variable breaks the uniformity of the code here, since all the other components explicitly declare the color hexcode. Could we change this back to the hex code or have all the others use the variable?

PR #871, opened 2018-08-13
Award Summary Blank submission date fix

kojodoesdesign:
https://federal-spending-transparency.atlassian.net/browse/DEV-590
- replaced empty submission date with fiscal year and quarter

### DIFF
--- a/src/js/models/v2/awards/financialSystemDetails/BaseFinancialSystemDetailsRow.js
+++ b/src/js/models/v2/awards/financialSystemDetails/BaseFinancialSystemDetailsRow.js
@@ -12,8 +12,10 @@ export const formatDate = (date) => date.format('MM/DD/YYYY');
 const BaseFinancialSystemDetailsRow = {
     populate(data) {
         this.id = data.financial_accounts_by_awards_id || '';
+        this.reporting_fiscal_year = data.submission.reporting_fiscal_year;
+        this.reporting_fiscal_quarter = data.submission.reporting_fiscal_quarter;
         this._submissionDate = (
-            (data.certified_date && parseDate(data.certified_date)) || null
+            (data.submission.reporting_fiscal_year && data.submission.reporting_fiscal_quarter) || null
         );
         this._fedAccountTitle = (data.treasury_account && data.treasury_account.federal_account
             && data.treasury_account.federal_account.account_title) || '';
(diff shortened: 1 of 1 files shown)


src/js/models/v2/awards/financialSystemDetails/BaseFinancialSystemDetailsRow.js
```diff
@@ -12,8 +12,10 @@ export const formatDate = (date) => date.format('MM/DD/YYYY');
 const BaseFinancialSystemDetailsRow = {
     populate(data) {
         this.id = data.financial_accounts_by_awards_id || '';
+        this.reporting_fiscal_year = data.submission.reporting_fiscal_year || null;
```
ebdabbs: Could you use came case variables to match the naming convention used in our models? I would also prefix both year and quarter with an underscore to indicate that they are not formatted/ only accessed by the getter.

src/js/models/v2/awards/financialSystemDetails/BaseFinancialSystemDetailsRow.js
```diff
@@ -12,8 +12,10 @@ export const formatDate = (date) => date.format('MM/DD/YYYY');
...
+        this.reporting_fiscal_quarter = data.submission.reporting_fiscal_quarter || null;
```
ebdabbs: First check for the existence of `data.submission` to prevent a `Cannot read property of undefined` error in case the submission object isn't present in the response. You can see examples of where we do this for `data.treasury_account`, etc. below.

src/js/models/v2/awards/financialSystemDetails/BaseFinancialSystemDetailsRow.js
```diff
@@ -12,8 +12,10 @@ export const formatDate = (date) => date.format('MM/DD/YYYY');
...
         this._submissionDate = (
-            (data.certified_date && parseDate(data.certified_date)) || null
+            (data.submission.reporting_fiscal_year && data.submission.reporting_fiscal_quarter) || null
```
ebdabbs: `this._submissionDate` is an unnecessary variable because you can check for the existence of `this._year` and `this._quarter` in the `submissionDate` getter function.

src/js/containers/award/table/FinancialSystemTableContainer.jsx
```diff
@@ -88,7 +88,8 @@ export class FinancialSystemTableContainer extends React.Component {
                     value: awardId
                 }
             ],
-            order: [`${sortDirection}${this.state.sort.field}`],
+            order: this.state.sort.field === 'submission__reporting_fiscal_year' ?
+                [`${sortDirection}${this.state.sort.field}`, `-submission__reporting_fiscal_quarter`] : [`${sortDirection}${this.state.sort.field}`],
```
ebdabbs: Use `sortDirection` instead of hard-coding `-` so that the quarters' ascending/descending direction will match the years'

PR #875, opened 2018-08-17
Added disclaimer on recipient landing page

kojodoesdesign:
https://federal-spending-transparency.atlassian.net/browse/DEV-1421

### DIFF
--- a/src/_scss/layouts/landingPage/landingPage.scss
+++ b/src/_scss/layouts/landingPage/landingPage.scss
@@ -19,7 +19,13 @@
             }
         }
     }
-		.landing-disclaimer {
+		.landing-page__disclaimer {
 			margin-top: rem(45);
+			.landing-page__note {
+				font-weight: bold;
+			}
+			.landing-page__entities {
+				font-style: italic;
+			}
 		}
 }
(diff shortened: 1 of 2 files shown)


src/js/components/recipientLanding/RecipientLandingContent.jsx
```diff
@@ -65,6 +65,9 @@ export default class RecipientLandingContent extends React.Component {
                     pageNumber={this.props.pageNumber}
                     totalItems={this.props.totalItems}
                     pageSize={this.props.pageSize} />
+                <p className="landing-disclaimer">
+                    <b>Note:</b> Profiles are not included for certain recipient names such as Multiple Recipients, Multiple Foreign Recipients, Miscellaneous Foreign Awardees, Private Individual, Individual Recipient, and Redacted Due to PII.
```
dtrinh50: Instead of using the <b> tag for styling: see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b, use CSS instead.

src/_scss/layouts/landingPage/landingPage.scss
```diff
@@ -19,4 +19,7 @@
             }
         }
     }
+		.landing-disclaimer {
```
dtrinh50: Follow BEM naming conventions, so this should be `.landing-page__disclaimer`.

PR #883, opened 2018-08-24
Temporarily disabling glossary link on parent recipient label

kojodoesdesign:
https://federal-spending-transparency.atlassian.net/browse/DEV-1479
Temporarily disabling parent recipient glossary link as per service desk request

### DIFF
--- a/src/js/components/recipient/RecipientOverview.jsx
+++ b/src/js/components/recipient/RecipientOverview.jsx
@@ -6,7 +6,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { CaretRight, Glossary } from 'components/sharedComponents/icons/Icons';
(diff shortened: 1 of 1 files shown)


src/js/components/recipient/RecipientOverview.jsx
```diff
@@ -43,11 +43,10 @@ export default class RecipientOverview extends React.Component {
         else if (recipient.level === 'P') {
             // This is a parent recipient
             label = (
-                <a
-                    href={`#/recipient/${this.props.recipient.id}/?glossary=parent-duns`}
+                <span
                     className="recipient-overview__label recipient-overview__label_parent">
                     Parent Recipient <Glossary />
```
ebdabbs: Could you remove the Glossary icon as well (since it is used to indicate a link to the glossary elsewhere on the site)?

PR #902, opened 2018-09-13
Award Summary pulling wrong description

kojodoesdesign:
https://federal-spending-transparency.atlassian.net/browse/DEV-934

### DIFF
--- a/src/js/components/award/contract/ContractDetails.jsx
+++ b/src/js/components/award/contract/ContractDetails.jsx
@@ -5,6 +5,9 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+
+import AwardDescription from 'containers/award/awards/AwardDescription';
+
 import * as TimeRangeHelper from 'helpers/timeRangeHelper';
 import * as SummaryPageHelper from 'helpers/summaryPageHelper';
 import { idvAwardTypes } from 'dataMapping/contracts/idvAwardTypes';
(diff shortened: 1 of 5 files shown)


src/js/components/award/contract/ContractDetails.jsx
```diff
@@ -37,9 +40,9 @@ export default class ContractDetails extends React.Component {
                         }} />
                     <table>
                         <tbody>
-                            <DetailRow
+                            <AwardDescription
+                                selectedAward={award}
                                 title="Description"
-                                value={award.description}
                                 overflow={award.description.length > maxChars} />
```
ebdabbs: The overflow prop should be calculated within the award description container using the length of the new description value.

src/js/containers/award/awards/AwardDescription.jsx
```diff
@@ -0,0 +1,90 @@
+/**
+  * AwardContainer.jsx
```
ebdabbs: Could you rename this file `AwardDescriptionContainer.jsx` to match naming convention? I would also rename the `awards` folder to `contracts` to match the components directory structure.

src/js/containers/award/contracts/AwardDescription.jsx
```diff
@@ -0,0 +1,90 @@
...
+            });
+    }
+
+
+    render() {
+        return (
+            <DetailRow
+                title={this.props.title}
+                value={this.state.lastDescription}
+                maxChars={this.state.lastDescription ? this.state.lastDescription.length > this.props.maxChars : ''} />
```
ebdabbs: Detail row takes a boolean `overflow` prop (https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/components/award/DetailRow.jsx#L14), to which you could just pass `this.state.lastDescription.length > this.props.maxChars`

PR #903, opened 2018-09-13
Fixed Recipient Landing Icons

dtrinh100:
- Fixed the icons styling for Recipient Landing page

- [ ] Code Review

### DIFF
--- a/src/_scss/pages/recipientLanding/recipientLandingPage.scss
+++ b/src/_scss/pages/recipientLanding/recipientLandingPage.scss
@@ -10,6 +10,31 @@
     @import "pages/search/results/table/_tableTypes";
     @import "pages/search/results/table/_resultsTablePicker";
 
+    .recipient-landing__icon {
+        color: $color-white;
+        margin-right: rem(10);
+        padding: 0 rem(4) 0 rem(4);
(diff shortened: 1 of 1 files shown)


src/_scss/pages/recipientLanding/recipientLandingPage.scss
```diff
@@ -10,6 +10,31 @@
...
+        border-radius: rem(3);
+        font-weight: 600;
+        font-size: rem(14);
+        display: inline-block;
+        width: rem(18);
+        height: rem(18);
+        text-align: center;
+
+        &.recipient-landing__icon_parent {
+            background-color: #205493;
```
kojodoesdesign: Could we use variables instead of the hex codes, to make it working with these classes easier in the future?

dtrinh100: I think using variables would make sense if we use the colors throughout the website, not sure if that's the case with this.

kojodoesdesign: Agreed, however we use these hex codes in our variables scss file, and the associated variables are used multiple times throughout the site.

PR #919, opened 2018-10-01
Hotfix Fiscal Year Incident

dtrinh50:
https://federal-spending-transparency.atlassian.net/browse/DEV-1626

- Fixes fiscal year incident on the spending explorer and federal accounts page.
- 2 automated tests were created to prevent a similar incident in the future.

### DIFF
--- a/src/js/containers/explorer/detail/helpers/explorerQuarters.js
+++ b/src/js/containers/explorer/detail/helpers/explorerQuarters.js
@@ -20,7 +20,7 @@ export const mostRecentQuarter = () => {
     let quarter = FiscalYearHelper.convertDateToQuarter(todayAdjusted);
     // use currentFiscalYear because we will perform the same logic (but in a more quarter-specific
     // manner) here as defaultFiscalYear
-    let year = FiscalYearHelper.currentFiscalYear();
+    let year = FiscalYearHelper.defaultFiscalYear();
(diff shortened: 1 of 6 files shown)


src/js/containers/explorer/detail/helpers/explorerQuarters.js
```diff
@@ -20,14 +20,13 @@ export const mostRecentQuarter = () => {
     let quarter = FiscalYearHelper.convertDateToQuarter(todayAdjusted);
     // use currentFiscalYear because we will perform the same logic (but in a more quarter-specific
```
ebdabbs: Could you remove this comment since we're now using `defaultFiscalYear`?

tests/containers/explorer/detail/helpers/explorerQuarters-test.js
```diff
@@ -60,6 +60,14 @@ describe('explorerQuarters', () => {
                 year: 1911
             });
         });
+        it('should use the previous fiscal year as the fiscal year on October 1st', () => {
```
ebdabbs: Can we generalize the description to `should use the previous fiscal year as the FY when the current date is between Oct 1 and Feb 14`? (You can still use Oct 1 as the test value). Similar for FY helper

PR #1027, opened 2019-03-15
Links to referenced awards table

ebdabbs:
**High level description:**

On IDV summary pages, implements
* two links that scroll to the Referenced Awards table
* referenced awards count in Related Awards section

**Technical details:**

* Passes the function `jumpToSection` to components that have the links
* Adds the Related Awards section to the list of sections that can be scrolled to with the `jumpToSection` function
* Passes the referenced award counts to the Related Awards section
* Includes some improvements to mobile-responsiveness
* Moves the award amounts API request to the IDV helper from the search helper

**JIRA Ticket:**
[DEV-2194](https://federal-spending-transparency.atlassian.net/browse/DEV-2194)

**Mockup:**
https://bahdigital.invisionapp.com/share/69IA8EPGPCM#/296002761_Award_Summary_2-0_-_IDV_-MVP-

- [ ] Code review
- [x] Design review


src/js/components/awardv2/AwardV2.jsx
```diff
@@ -35,6 +35,10 @@ const awardSections = [
     {
         section: 'additional-information',
         label: 'Additional Information'
+    },
+    {
+        section: 'referenced-awards',
+        label: 'Referenced Awards'
     }
 ];
```
MaxwellKendall: Again, out of scope for this PR, but would love to know your thoughts about the `jumpToSection` method on this class regarding direct interaction with the DOM.

src/js/components/awardv2/visualizations/amounts/AggregatedAwardAmountsInfo.jsx
```diff
@@ -60,16 +60,16 @@ export default class AggregatedAwardAmountsInfo extends React.Component {
                 <div className="award-amounts__data">
                     <span>Awards that Reference this IDV</span><span>{awardAmounts.idvCount + awardAmounts.contractCount}</span>
                 </div>
-                <a
-                    href="/"
-                    className="award-viz__link">
-                    <div className="award-viz__link-icon award-viz__link-icon_hidden">
+                <button
+                    onClick={() => this.props.jumpToSection('referenced-awards')}
+                    className="award-viz__button">
```
MaxwellKendall: Another convo might be worth having around the best practice for passing `fns` to `onClick`. This is typically what I do as well when passing a `fn` with a certain parameter, but [the docs](https://reactjs.org/docs/faq-functions.html) indicate there are performance implications to doing this.

src/js/components/awardv2/visualizations/amounts/AwardAmounts.jsx
```diff
@@ -45,7 +46,14 @@ export default class AwardAmounts extends React.Component {
...
-        const content = this.state.active === 'awards' ? (<AwardAmountsContainer awardId={this.props.awardId} />) : (<IDVAmounts awards={awards} />);
+        const content = this.state.active === 'awards' ? (
+            <AwardAmountsContainer
+                jumpToSection={this.props.jumpToSection}
+                awardId={this.props.awardId} />
+        ) : (
+            <IDVAmounts
+                awards={awards} />
+        );
         return (
```
MaxwellKendall: Very clean ternary. Love it!! 💯 

Also wondering in principle, if we're rendering a container, should we be passing props from the parent component if the value of the prop is derived directly from redux?

I guess this would be a question around principle which may not really matter to us, but I'm curious if we'd like to develop a consensus for this scenario as well?

My gut is container components should refer to redux directly via `connect` if possible.

src/js/components/awardv2/visualizations/overview/RelatedAwards.jsx
```diff
@@ -33,6 +35,16 @@ export default class RelatedAwards extends React.Component {
                     </div>
                     {parentLink}
                 </div>
+                <button
+                    onClick={() => this.props.jumpToSection('referenced-awards')}
+                    className="award-viz__button">
```
MaxwellKendall: Same question as above RE: passing anonymous `fns` to `onClick`

src/js/redux/actions/awardV2/awardActions.js
```diff
@@ -12,6 +12,11 @@ export const setCounts = (state) => ({
     counts: state
 });
 
+export const setAggregatedAmounts = (state) => ({
+    type: 'SET_AMOUNTS',
+    amounts: state
+});
+
```
MaxwellKendall: Could you explain why we're calling the params to most actions `state`? I think this is just a different way of thinking about it which I haven't really understood very well as of yet.

src/js/components/awardv2/idv/IdvContent.jsx
```diff
@@ -20,6 +20,7 @@ import { AWARD_V2_OVERVIEW_PROPS } from '../../../propTypes';
 
 const propTypes = {
     awardId: PropTypes.string,
+    counts: PropTypes.object,
     overview: AWARD_V2_OVERVIEW_PROPS,
```
MaxwellKendall: We could put a TODO here to add prop types for this object, or just define it here with `shape()` or put it in the `/prop-types` dir

src/js/components/awardv2/visualizations/amounts/AwardsBanner.jsx
```diff
@@ -25,7 +29,16 @@ export default class AwardsBanner extends React.Component {
 
             <div className={`award-amounts__banner ${!this.state.toggle ? "award-amounts__banner_hidden" : ""}`}>
                 <span className="award-amounts__banner-info-icon"><InfoCircle /></span>
-                <p>The information in this tab is pulled from the combined data of awards that reference this IDV, not the IDV itself. To see those awards, scroll to the <a href="/">referencing awards table</a> on this page.</p>
+                <p>The information in this tab is pulled from the combined data of awards that reference this IDV, not the IDV itself. To see those awards, scroll to the
+                    &nbsp;
+                    <button
+                        onClick={() => this.props.jumpToSection('referenced-awards')}
+                        className="award-viz__button">
```
MaxwellKendall: Same q as above RE: passing anonymous fns to onClick

src/js/containers/awardV2/visualization/AwardAmountsContainer.jsx
```diff
@@ -99,16 +100,21 @@ export class AwardAmountsContainer extends React.Component {
         });
 
         // Store the counts in Redux for use in the referenced awards table
+        // and related awards section
         this.props.setCounts({
             idvs: data.idv_count,
-            contracts: data.contract_count
+            contracts: data.contract_count,
+            total: data.idv_count + data.contract_count
         });
```
MaxwellKendall: Looks like we're putting into redux an object with a different shape than what exists in the model.

Is that true for what we currently have prior to the delivery of this PR?

Award Model Object !== `redux.award`?

src/js/components/awardv2/visualizations/amounts/AggregatedAwardAmountsInfo.jsx
ebdabbs: Good call, removed arrow functions from render

src/js/components/awardv2/visualizations/amounts/AwardAmounts.jsx
ebdabbs: Yep, changed `AwardAmountsContainer` to get the award id from Redux.

src/js/redux/actions/awardV2/awardActions.js
ebdabbs: That's a good question, and I think it's just a convention referring to the application "state" (which gets confusing for the state profile page reducer 😂). But I think you're right that it could be more clear to just call the parameter `amounts` instead of `state`, which is how they show examples in the [Redux docs](https://redux.js.org/basics/actions#action-creators)

src/js/components/awardv2/idv/IdvContent.jsx
ebdabbs: Added to `/prop-types` dir

PR #1030, opened 2019-03-21
Release IDV summary pages

ebdabbs:
**High level description:**

* `/#/award/{id}` urls will now display the v2 summary pages for IDVs
* `/#/award/{id}` urls still display the v1 summary pages for non-IDVs
* `/#/award_v2/{id}` can still be used to preview v2 summary pages for non-IDVs

**Technical details:**

The `/award` route now points to `AwardV2Container`.

* The `AwardV2Container` makes an API call to the `v2/awards` endpoint, and uses the response to determine if `category === 'idv'`
* `AwardV2Container` checks the current url to see if it contains `'award_v2'`
* If the current award is either an IDV or we are at an `/award_v2/` url, show the v2 Award Component
* If it is not an IDV and we are at a normal `/award/` url, render the v1 `AwardContainer`, which makes an API call to the `v1/awards` endpoint and renders subsequent v1 components as before

Note: `awardId` is passed as a param to the v1 `AwardContainer` because it requires the internal id format (e.g. `68935929`), which is not yet available via Redux since v2 uses the generated unique id format (e.g. `CONT_AW_8000_-NONE-_NNK14MA74C_-NONE-`)

**JIRA Ticket:**
[DEV-2233](https://federal-spending-transparency.atlassian.net/browse/DEV-2233)

**Mockup:**
N/A

- [x] Code review

### DIFF
--- a/src/js/containers/awardV2/AwardV2Container.jsx
+++ b/src/js/containers/awardV2/AwardV2Container.jsx
@@ -130,7 +130,6 @@ export class AwardContainer extends React.Component {
             if (this.props.award.category === 'idv' || isV2url) {
                 content = (<Award
                     award={this.props.award}
-                    inFlight={this.state.inFlight}
                     noAward={this.state.noAward} />);
             }
             else {
--- a/src/js/containers/awardV2/AwardV2Container.jsx
+++ b/src/js/containers/awardV2/AwardV2Container.jsx
@@ -130,7 +130,6 @@ export class AwardContainer extends React.Component {
             if (this.props.award.category === 'idv' || isV2url) {
                 content = (<Award
                     award={this.props.award}
-                    awardId={this.props.params.awardId}
                     inFlight={this.state.inFlight}
                     noAward={this.state.noAward} />);
             }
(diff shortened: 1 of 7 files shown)


src/js/containers/router/RouterRoutes.jsx
```diff
@@ -61,7 +61,7 @@ const routes = {
             parent: '/award',
             component: (cb) => {
                 require.ensure([], (require) => {
-                    cb(require('containers/award/AwardContainer').default);
+                    cb(require('containers/awardV2/AwardV2Container').default);
                 });
```
MaxwellKendall: Blown away we have our own custom router... Really interested by that and impressed! Would love to hear the rationale around that sometime!

src/js/containers/awardV2/AwardV2Container.jsx
```diff
@@ -122,13 +124,20 @@ export class AwardContainer extends React.Component {
...
+                    award={this.props.award}
+                    inFlight={this.state.inFlight}
+                    noAward={this.state.noAward} />);
+            }
+            else {
+                content = (<AwardV1Container awardId={this.props.params.awardId} />);
+            }
+        }
+        return content;
     }
```
MaxwellKendall: Do we really need `line 129` as a condition? I don't see where we're handling in the UI the case where `inFlight` is true. 

Should we be handling that here in the container, or letting the child components handle it? It appears `<Award />` wants to do something with it, but with these changes it will always be `true`.

PR #1031, opened 2019-03-22
[DEV-2289] IDV ToolTips for Multiple Sections on Summary Page

MaxwellKendall:
**High level description:**

New Component implemented in multiple sections of IDV Summary Page.

**Technical details:**

New Component uses offSet properties to derive appropriate position on screen relative to the <body> element. 

One button type (`.table-type-toggle`) had to be refactored away from using `position` & `top` CSS properties because it was interfering with ability to derive appropriate offSet values

**JIRA Ticket:**
[DEV-2289](https://federal-spending-transparency.atlassian.net/browse/DEV-2289)

**Commits:**
* 79c9608 -- using `margin-bottom` instead of `position relative` and `top` to prevent issue with deriving offset properties
~62b536c -- moving `closeToolTip()` to `mouseLeave` toolTipContent instead of `toolTipButton` so that users can scroll to see toolTip content when it is longer than view port~ (31240cc -- puts the previous UX back where tool tip always closes onMouseLeave from icon)
dcf88fd -- implementing new prop "wide" to make tool tip wider so scrolling is not necessary with longer tool tip content
a057812 & 04faae6 -- New ToolTip & toolTip Content content (Awards under this IDV)
f17be3b -- adding new styles for transaction history tool tip because its so long

**Mockup:**
`N/A` (?) Couldn't find any on the JIRA ticket

**Screenshots:**
image

image

- [x] Code review
- [ ] API contract approved by @backendDev (if applicable) `N/A`
- [ ] [API #1234](https://github.com/fedspendingtransparency/usaspending-api/pull/1234) merged (if applicable) `N/A`


src/js/components/awardv2/idv/InfoTooltip.jsx
```diff
@@ -93,18 +96,21 @@ export default class InfoTooltip extends React.Component {
...
+                    }}>
+                    <div
+                        role="button"
+                        tabIndex="0"
                         onBlur={this.closeTooltip}
                         className="award__icon"
                         onFocus={this.showTooltip}
+                        onKeyPress={this.showTooltip}
                         onMouseEnter={this.showTooltip}
-                        onMouseLeave={this.closeTooltip}
```
jonhill13: My main concern with moving the `onMouseLeave` from the icon to the tooltip removes the functionality of removing the tooltip when I leave the icon. With these changes I will now have to enter the tooltip div and then leave that div to remove the tooltip.

jonhill13:
My only other concerns are these two points from the [Jira ticket](https://federal-spending-transparency.atlassian.net/browse/DEV-2289)

• Awards that Reference this IDV (can be pushed to follow-up ticket as needed since added late)
3. Remove extra word "Parent" in the related awards tooltip (second bullet should start with "Award Orders Under this IDV")

src/js/components/awardv2/idv/InfoTooltip.jsx
MaxwellKendall: Yep I changed this back actually.

PR #1041, opened 2019-04-01
Normal scenario and insufficient data

ebdabbs:
**High level description:**

Implements an "Insufficient Data" message for edge cases in the IDV aggregate Award Amounts visualization. 

**Technical details:**

- Refactored names used in the models to reflect updated terminology in the visualization
- Added a helper function that, given award amounts, determines whether they indicate
  - Normal
  - Overspend
  - Extreme overspend
  - Everything else
- Perform a check for the normal situation and implement the "Insufficient Data" message as the fallback. Subsequent PRs will check for "Overspend" and "Extreme overspend" and implement their corresponding visualizations.
- Award id for an example of the "Insufficient Data" message in dev: `68860659` (overspend)

**JIRA Ticket:**
[DEV-2356](https://federal-spending-transparency.atlassian.net/browse/DEV-2356), subtask for step 2 of [DEV-2224](https://federal-spending-transparency.atlassian.net/browse/DEV-2224)

**Mockup:**
https://bahdigital.invisionapp.com/share/69IA8EPGPCM#/296009664_combinedAwardsSection--unavailable

- [x] Code review
- [x] Design review

### DIFF
--- a/src/js/models/v2/awardsV2/BaseAwardAmounts.js
+++ b/src/js/models/v2/awardsV2/BaseAwardAmounts.js
@@ -29,10 +29,15 @@ const BaseAwardAmounts = {
         return MoneyFormatter.formatMoneyWithPrecision(this._obligation, 2);
     },
     get obligationFormatted() {
-        // TODO - Lizzie: handle negative values
-        if (this._obligation >= MoneyFormatter.unitValues.MILLION) {
+        if (Math.abs(this._obligation) >= MoneyFormatter.unitValues.MILLION) {
             const units = MoneyFormatter.calculateUnitForSingleValue(this._obligation);
+            if (this._obligation < 0) {
+                return `(${MoneyFormatter.formatMoneyWithPrecision(Math.abs(this._obligation) / units.unit, 1)} ${units.unitLabel})`;
+            }
             return `${MoneyFormatter.formatMoneyWithPrecision(this._obligation / units.unit, 1)} ${units.unitLabel}`;
+        }
+        else if (this._obligation < 0) {
+            return `(${Math.abs(MoneyFormatter.formatMoney(this._obligation))})`;
         }
         return MoneyFormatter.formatMoney(this._obligation);
     },
--- a/src/js/models/v2/awardsV2/BaseAwardAmounts.js
+++ b/src/js/models/v2/awardsV2/BaseAwardAmounts.js
@@ -3,7 +3,6 @@
  * Created by David Trinh 12/19/18
  */
 
-
 import * as MoneyFormatter from 'helpers/moneyFormatter';
 
 const BaseAwardAmounts = {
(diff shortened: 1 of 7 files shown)


src/js/models/v2/awardsV2/BaseAwardAmounts.js
```diff
@@ -29,11 +29,16 @@ const BaseAwardAmounts = {
         return MoneyFormatter.formatMoneyWithPrecision(this._obligation, 2);
     },
     get obligationFormatted() {
-        // TODO - Lizzie: handle negative values
-        if (this._obligation >= MoneyFormatter.unitValues.MILLION) {
+        if (Math.abs(this._obligation) >= MoneyFormatter.unitValues.MILLION) {
             const units = MoneyFormatter.calculateUnitForSingleValue(this._obligation);
+            if (this._obligation < 0) {
+                return `(${MoneyFormatter.formatMoneyWithPrecision(Math.abs(this._obligation) / units.unit, 1)} ${units.unitLabel})`;
```
MaxwellKendall: Could you walk me through what's happening here for my future reference? Specifically how this function handles the first parameter `absolute value / Million`?

MaxwellKendall: Also I see we're not returning a negative value here. I guess there is a check for if the `.obligationFormatted()` should be negative elsewhere.

Would like to hear your thought process around this choice as to why the return value of this function is absolute!

tests/helpers/aggregatedAmountsHelper-text.js
```diff
@@ -0,0 +1,60 @@
...
+        });
+
+        it('should return "insufficientData" for negative obligations', () => {
+            const mockAmounts = {
+                _obligation: -55,
+                _combinedCurrentAwardAmounts: 75,
+                _combinedPotentialAwardAmounts: 100
+            };
+            const mockedScenario = AggregatedAmountsHelper.determineScenario(mockAmounts);
+            expect(mockedScenario).toEqual("insufficientData");
```
MaxwellKendall: I think this maybe answers my question elsewhere regarding handling negative obligations and you're gonna iterate on this to handle the negative scenario?

Not as familiar as perhaps I should be with the relevant conversations around the scope of this work.

src/js/models/v2/awardsV2/BaseAwardAmounts.js
ebdabbs: Good question - the [mockup for de-obligated](https://bahdigital.invisionapp.com/share/69IA8EPGPCM#/296009661_combinedAwardsSection--deobligation) scenario is displaying negative values that way. And the width of that diagonal stripe section will represent the absolute value of the negative obligations. The actual/negative value will also still be available via the `_obligation` property.

MaxwellKendall: Sounds good.

PR #1047, opened 2019-04-05
[DEV-2226] A/C 1 and 2: Removing IDV_B from subsection search and Calls from BPA Description

MaxwellKendall:
**High level description:**
* Removing `IDV_B` award type from Advanced Search award type sub-section filter area
* Omitting "Calls" from the `BPA` award type sub-section filter check-box label

**Technical details:**
Instead of changing the relevant data maps (which are very widely used), I'm changing the relevant values after they've been imported as stored as a variable in the relevant component, so that `IDV_B` is not included in the `awardTypeGroups` under the IDV key-space, and the "BPA" awardTypeCode value omits "Calls."

**JIRA Ticket:**
[DEV-2226](https://federal-spending-transparency.atlassian.net/browse/DEV-2226) (A/C 1 &2)

**Commits:**
4c5ed1c -- Removing IDV_B award type from Award Type Advanced Search Filter sub-section (e1c50e2 removes a propType that was not used)
b8508b8 -- Omitting "Calls" from BPA Award type description (8a565eb uses `Object.assign() instead of `...`)

- [ ] Code review
- [ ] [API #1731](https://github.com/fedspendingtransparency/usaspending-api/pull/1731) merged (if applicable) (Front end doesn't depend on this, but (A/C 3 &4 do)

### DIFF
--- a/src/js/components/search/filters/awardType/AwardType.jsx
+++ b/src/js/components/search/filters/awardType/AwardType.jsx
@@ -45,10 +45,9 @@ const awardTypesData = [
     }
 ];
 
-const awardTypeCodesData = {
-    ...awardTypeCodes,
+const awardTypeCodesData = Object.assign(awardTypeCodes, {
     IDV_E: "Blanket Purchase Agreements (BPA)"
-};
+});
 
 const propTypes = {
     awardTypes: PropTypes.arrayOf(PropTypes.object),
--- a/src/js/components/search/filters/awardType/AwardType.jsx
+++ b/src/js/components/search/filters/awardType/AwardType.jsx
@@ -10,41 +10,40 @@ import { awardTypeGroups, awardTypeCodes } from 'dataMapping/search/awardType';
 import PrimaryCheckboxType from 'components/sharedComponents/checkbox/PrimaryCheckboxType';
 import SubmitHint from 'components/sharedComponents/filterSidebar/SubmitHint';
 
-const defaultProps = {
-    awardTypes: [
-        {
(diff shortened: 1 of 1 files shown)


src/js/components/search/filters/awardType/AwardType.jsx
```diff
@@ -10,40 +10,42 @@ import { awardTypeGroups, awardTypeCodes } from 'dataMapping/search/awardType';
...
+        filters: awardTypeGroups.loans
+    },
+    {
+        id: 'award-other',
+        name: 'Other',
+        filters: awardTypeGroups.other
+    }
+];
+
+const awardTypeCodesData = Object.assign(awardTypeCodes, {
```
ebdabbs: You could also update the data mapping [here](https://github.com/fedspendingtransparency/usaspending-website/blob/DEV-2226_IDV_B/src/js/dataMapping/search/awardType.js#L25) instead of using `Object.assign`

MaxwellKendall: @ebdabbs ,

Are we sure that it's okay to update it there? Not sure if we want to remove this everywhere, or if removing it from here will remove it everywhere in the app...?

ebdabbs: @MaxwellKendall yep, in this case we do want to update it everywhere since without `Calls` is the more correct terminology for IDVs.

MaxwellKendall: Alright!!

PR #1057, opened 2019-04-15
[DEV-2409] Adding download button on the IDV Summary Page

MaxwellKendall:
### DNM until back end work associated with [this story](https://federal-spending-transparency.atlassian.net/browse/DEV-2336) has been complete.

**High level description:**
Adds a download button to the IDV Summary Page.

**Technical details:**
This PR implements an existing UX from the Keyword Search page on the new IDV Summary Page. It makes use of the existing components, styles, and action creators, while introducing a new end point that is still under development.

Irrelevant commits exist due to confusion around adding a modal and seeking to compose action creators together. For the former, I later learned the UX didn't require a modal. For the later, I learned we needed to add redux-middleware as a new dependency in order to achieve this implementation.

**JIRA Ticket:**
[DEV-2409](https://federal-spending-transparency.atlassian.net/browse/DEV-2409)

**Mockup:**
https://bahdigital.invisionapp.com/share/69IA8EPGPCM#/screens/296002761

**Commits:**
* 561392e & d5b7637-- styles for download button
* 903fff3 & 135a7c1 -- declaring and using new helper
* 0f29fbe -- tests

- [x] Code review
- [x] Revert base url for new helper to point to a global constant (currently pointing at mock)
- [x] API contract approved by @backendDev [link](https://github.com/fedspendingtransparency/usaspending-api/pull/1751)
- [x] [API #2336](https://github.com/fedspendingtransparency/usaspending-api/pull/2336) merged (if applicable) `not opened yet`


src/js/containers/awardV2/AwardV2Container.jsx
```diff
@@ -144,6 +188,12 @@ export class AwardContainer extends React.Component {
 AwardContainer.propTypes = propTypes;
 
 export default connect(
-    (state) => ({ award: state.awardV2 }),
-    (dispatch) => bindActionCreators(awardActions, dispatch)
+    (state) => ({ award: state.awardV2, isDownloadPending: state.bulkDownload.download.pendingDownload }),
+    (dispatch) => ({
+        setDownloadExpectedUrl: (url) => dispatch(setDownloadExpectedUrl(url)),
```
ebdabbs: Any rationale for removing `bindActionCreators`? We've also use a pattern of creating a `combinedActions` object that gets passed to `bindActionCreators` in other containers (e.g. [here](https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/containers/search/SearchSidebarSubmitContainer.jsx#L123))

MaxwellKendall: Mainly out of convenience. Was having some trouble cause I'm not using an alias when importing the AC's and Im not sure how elegant I could use `Object.assign` to combine all of them?

src/js/components/search/header/DownloadButton.jsx
```diff
@@ -41,12 +41,10 @@ export default class DownloadButton extends React.Component {
     }
 
     onClick(e) {
-        if (!this.props.downloadAvailable || this.props.downloadInFlight) {
-            e.preventDefault();
-            return;
+        e.preventDefault();
+        if (this.props.downloadAvailable || !this.props.downloadInFlight) {
```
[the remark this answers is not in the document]
MaxwellKendall: I don't think so. It was an `| |` and I think it was intended to exit the function in _either_ case.

src/js/containers/awardV2/AwardV2Container.jsx
MaxwellKendall: Also was getting this error when trying to use it in the same way as the docs [here](https://react-redux.js.org/using-react-redux/connect-mapdispatch).

image

I think we should try and squeeze in some babel config so we can use the spread operator. The benefit of this would be instead of doing `Object.assign()` which requires the parameters to be objects (?), we can use the spread operator which is a bit easier IMO. Really don't understand why it's not already in there based on the current config... 

The error above is about using `redux-thunk` which would be a super small config change and a new dependency, but would allow for some more flexibility around implementing different patterns etc...

PR #1060, opened 2019-04-16
Add IDV Table to Keyword Search Page

jonhill13:
**High level description:**

Add IDV table to Keyword Search Page.

**Technical details:**
- add idvs to table types
- add columns for idvs to keyword search data mapping
- add default column for idvs
- remove keyword search variable from route

**JIRA Ticket:**
[DEV-1807](https://federal-spending-transparency.atlassian.net/browse/DEV-1807)

**Mockup:**
TBD

- [ ] Code review
- [x] Design review
- [x] [API #TBD](https://github.com/fedspendingtransparency/usaspending-api/pull/TBD) merged 

### DIFF
--- a/src/js/containers/keyword/table/ResultsTableContainer.jsx
+++ b/src/js/containers/keyword/table/ResultsTableContainer.jsx
@@ -293,17 +293,18 @@ export default class ResultsTableContainer extends React.Component {
         const columns = {};
         tableTypes.forEach((type) => {
             const allColumns = map(availableColumns(type.internal), (data) => data.title);
-
-            const parsedColumns = {};
-            availableColumns(type.internal).forEach((data) => {
-                parsedColumns[data.title] =
+            const parsedColumns = availableColumns(type.internal).reduce((result, data) => Object.assign(
+                {},
+                result,
                 {
-                    columnName: data.title,
-                    displayName: data.displayName || data.title,
-                    width: measureTableHeader(data.displayName || data.title),
-                    defaultDirection: 'desc'
-                };
-            });
+                    [data.title]: {
+                        columnName: data.title,
+                        displayName: data.displayName || data.title,
+                        width: measureTableHeader(data.displayName || data.title),
+                        defaultDirection: 'desc'
+                    }
+                }),
+            {});
 
             columns[type.internal] = {
                 visibleOrder: allColumns,
--- a/src/js/containers/keyword/table/ResultsTableContainer.jsx
+++ b/src/js/containers/keyword/table/ResultsTableContainer.jsx
@@ -6,7 +6,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { isCancel } from 'axios';
-import { uniqueId } from 'lodash';
+import { uniqueId, map } from 'lodash';
 import * as KeywordHelper from 'helpers/keywordHelper';
 import { availableColumns, defaultSort } from 'dataMapping/keyword/resultsTableColumns';
 import { awardTypeGroups } from 'dataMapping/search/awardType';
@@ -176,7 +176,7 @@ export default class ResultsTableContainer extends React.Component {
         }
         const resultLimit = 35;
 
-        const requestFields = availableColumns(this.state.tableType);
+        const requestFields = map(availableColumns(this.state.tableType), (data) => data.title);
         const tableType = this.state.tableType;
 
         const params = {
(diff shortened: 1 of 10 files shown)


src/js/dataMapping/navigation/menuOptions.js
```diff
@@ -14,7 +14,7 @@ export const searchOptions = [
     {
         label: 'Keyword Search',
         url: '#/keyword_search',
-        enabled: kGlobalConstants.KEYWORD_AVAILABLE
+        enabled: true
     }
```
MaxwellKendall: @jonhill13 I too like to live dangerously...........

jonhill13: LOL, we will need to delete that variable from dev, test, and prod config. But I thought since keyword search is live, we should live bold and `true`

src/js/containers/search/table/ResultsTableContainer.jsx
```diff
@@ -132,10 +132,10 @@ export class ResultsTableContainer extends React.Component {
         // load every possible table column up front, so we don't need to deal with this when
         // switching tabs
         const columns = tableTypes.concat(subTypes).reduce((cols, type) => {
-            const visibleColumns = defaultColumns(type.internal);
+            const visibleColumns = map(defaultColumns(type.internal), (data) => data.title);
             const parsedColumns = {};
-            visibleColumns.forEach((title) => {
-                parsedColumns[title] = this.createColumn(title);
+            defaultColumns(type.internal).forEach((data) => {
+                parsedColumns[data.title] = this.createColumn(data.displayName, data.title);
```
MaxwellKendall: I think we should use es6 methods as much as possible rather than lodash, in this case then we could turn line 137 to a `.reduce` and line 135 to use `.map`

```javascript
// line 135
visibleColumns = defaultColumns(type.internal).map((col) => col.title);
// line 136 + 137
parsedColumns = defaultColumns(type.internal).reduce((parsedCols, data) => {
    return Object.assign({}, parsedCols, {
        [data.title]: this.createColumn(data.displayName, data.title)
    });
}, {});
```

MaxwellKendall: Couldnt line 138 also be   `this.createColumn(data.displayName || data.title)`? That way we could just use one param in this.createColumn

jonhill13: Yeah, I do not really care JS map or reduce v. Lodash map or reduce to simply map or reduce an array. The reasons I use Lodash are because I used it heavily on my last project so I am extremely comfortable with it. We use it in a lot of places in this code base. There are a lot of useful functions that Lodash has already, that I also could write myself with JS map and JS reduce, and other JS methods, but since Lodash already has them, this saves me time and provides standardization without having to create our own.

The `forEach` for `parsedColumns` is a good catch, but we should just use map. Not sure why to use reduce?

The `createColumn` function needs two fields because it is setting the `columnName` and a `displayName` property among others for the column data set. The `columnName` is more of a key for the data sent to the API and the `displayName` is what the user sees.  When clicking sort and probably other functionality the table sends the `columnName`. If a column had a `displayName` and we set the `columnName` to the `displayName` we would get an error from the API.

tony-sappe:
@jonhill13 Backend PR was merged. However the ES cluster hasn't been reindexed yet, so it isn't completely functional yet. I'll run it over the weekend

src/js/containers/search/table/ResultsTableContainer.jsx
MaxwellKendall: @jonhill13 ,

You're returning an object in the `.forEach` loop. In ES6 `.map` always returns an array, so in this case that's why you'd want to use `.reduce`. I'm good w/ you using `lodash` but I'm gonna use `ES6`, maybe this way we can both introduce one another to the various benefits of each. I personally don't have a ton of experience with lodash and would definitely be interested in learning while I review your PRs.

Understood regarding the `createColumn` fn. Thanks for the clarification, I see that now you are using only `title` for the `columnName` property.

jonhill13: Yeah, I like that using reduce to return and object instead of creating one before the forEach.

MaxwellKendall: 👌

src/js/containers/keyword/table/ResultsTableContainer.jsx
```diff
@@ -288,15 +292,15 @@ export default class ResultsTableContainer extends React.Component {
...
         const columns = {};
         tableTypes.forEach((type) => {
-            const allColumns = availableColumns(type.internal);
+            const allColumns = map(availableColumns(type.internal), (data) => data.title);
 
             const parsedColumns = {};
-            allColumns.forEach((title) => {
-                parsedColumns[title] =
+            availableColumns(type.internal).forEach((data) => {
+                parsedColumns[data.title] =
```
MaxwellKendall: could do the same thing here as in line 134 of the other `ResultsTableCtonainer` (use `.reduce`)

PR #1067, opened 2019-04-23
[dev-2458] Quick fix for wide tooltips that were not showing properly on full width components

MaxwellKendall:
**High level description:**
Adds new condition to handle edge case for tooltips -- when prop `wide` is true and parent component is full width

**Technical details:**
Adds a condition to see if the `window.innerWidth - tooltip.offsetLeft` is lower than anticipated (lower than 1/6 of the screen) to handle edge case of when tool tip is on a full width component and has the wide property set to true

**JIRA Ticket:**
[DEV-2458](https://federal-spending-transparency.atlassian.net/browse/DEV-2458)

- [x] Code review

### DIFF
--- a/src/js/components/awardv2/idv/InfoTooltip.jsx
+++ b/src/js/components/awardv2/idv/InfoTooltip.jsx
@@ -51,7 +51,7 @@ export default class InfoTooltip extends React.Component {
 
     closeTooltip() {
         this.setState({
-            showInfoTooltip: false
+            showInfoTooltip: true
         });
     }
 
@@ -59,7 +59,14 @@ export default class InfoTooltip extends React.Component {
         const targetElement = this.referenceDiv;
         const offsetTop = targetElement.offsetTop - 15;
         let tooltipWidth = 375;
-        if (this.props.wide) {
+        // is the t.t. in a section that takes up the full width of the screen?
+        const isTooltipJustifiedRight = (window.innerWidth - targetElement.offsetLeft) < window.innerWidth / 6;
+
+        if (this.props.wide && isTooltipJustifiedRight) {
+            tooltipWidth = 700;
+        }
+        else if (this.props.wide) {
+            // is there at least 700px of space to the right for the t.t.?
             tooltipWidth = (window.innerWidth - targetElement.offsetLeft > 700)
                 ? 700
                 : window.innerWidth - targetElement.offsetLeft - 100;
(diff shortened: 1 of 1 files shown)


src/js/components/awardv2/idv/InfoTooltip.jsx
```diff
@@ -51,15 +51,22 @@ export default class InfoTooltip extends React.Component {
 
     closeTooltip() {
         this.setState({
-            showInfoTooltip: false
+            showInfoTooltip: true
```
ebdabbs: `closeTooltip` should continue to hide the tooltip - this is causing them to stay open indefinitely.

src/js/components/awardv2/idv/InfoTooltip.jsx
```diff
@@ -51,15 +51,22 @@ export default class InfoTooltip extends React.Component {
...
         });
     }
 
     measureOffset() {
         const targetElement = this.referenceDiv;
         const offsetTop = targetElement.offsetTop - 15;
         let tooltipWidth = 375;
-        if (this.props.wide) {
+        // is the t.t. in a section that takes up the full width of the screen?
+        const isTooltipJustifiedRight = (window.innerWidth - targetElement.offsetLeft) < window.innerWidth / 6;
```
ebdabbs: Not sure I'm following the logic here - why is offset compared to `window.innerWidth / 6`?

MaxwellKendall: @ebdabbs, 

The scenario I'm seeking to recognize is:

- Tooltip is on a full-width component
- Tooltip is invoked w/ the prop `wide`

In this case, the pixel value of `offsetLeft` for the tooltip is very near the value of the `window.innerWidth`. On my laptop, `window.innerWidth - offsetLeft` is ~90px whereas on my monitor it's about 240px. In either case, it's approximately less than `1/6` of `window.innerWidth`.

So, to answer your question, it's somewhat arbitrary, but this value was reliable during my testing. During my testing, I was able to confirm that `window.innerWidth - offsetLeft` is larger than `1/6` of the screen unless the conditions of this scenario, stated above, apply.

That said, my only real goal here was to not use a fixed width, given the variety of screen sizes that will affect the value of `window.innerWidth`.

src/js/components/awardv2/idv/InfoTooltip.jsx
MaxwellKendall: Yes I forgot to change this back! Good catch.

PR #1106, opened 2019-05-22
Pushing City Search to Staging

MaxwellKendall:


MaxwellKendall:
Not sure why this PR includes all the same commits from #1105. Should've only been commit b9f8868.

PR #1142, opened 2019-06-25
Dev 2646 IDV Activity Chart Tooltips

jonhill13:
**High level description:**

Add tooltips to IDV Activity Chart Bars

**JIRA Ticket:**
[DEV-2646](https://federal-spending-transparency.atlassian.net/browse/DEV-2646)

**Mockup:**
https://bahdigital.invisionapp.com/share/ACIAA6A938E#/screens/296010428

- [x] Code review
- [x] Design review @AGuyNamedMarco 


src/js/components/awardv2/idv/activity/ActivityChartTooltip.jsx
```diff
@@ -0,0 +1,286 @@
...
+                transform: ''
+            },
+            windowWidth: 0,
+            windowHeight: 0,
+            awardingAgency: props.data.awardingAgencyName,
+            recipient: props.data.recipientName,
+            truncated: false
+        };
+
+        this.measureWindow = this.measureWindow.bind(this);
```
MaxwellKendall: Should we debounce/throttle this fn?

src/js/components/awardv2/idv/activity/ActivityChartTooltip.jsx
```diff
@@ -0,0 +1,301 @@
+/**
+ * FederalAccountsTreeTooltip.jsx
+ * Created by Jonathan Hill 5/2/19
+ */
```
MaxwellKendall: update this?

src/js/components/awardv2/idv/activity/ActivityChartTooltip.jsx
```diff
@@ -80,8 +80,8 @@ export default class IdvActivityTooltip extends React.Component {
     }
 
     truncateText(awardingAgencyNameDiv, recipientNameDiv, theTooltipWidth) {
-        const { graphWidth, recipientName, awardingAgencyName } = this.props.data;
-        if (this.props.data.graphWidth < theTooltipWidth) {
+        return new Promise((resolve) => {
```
MaxwellKendall: Tested this locally and it appears we don't need to return a promise, we can just use the async / await on line 159. `setState` itself is async so I don't think we need to wrap all this in a promise. Maybe I'm missing something but that's my understanding.

jonhill13: Not sure we really want to do that considering https://stackoverflow.com/questions/47019199/why-does-async-await-work-with-react-setstate

PR #1161, opened 2019-07-08
Dev-3025 Add Note to IDV Activity Chart

jonhill13:
**High level description:**

Add Note to IDV Activity Chart for Edge Cases

**JIRA Ticket:**
[DEV-3025](https://federal-spending-transparency.atlassian.net/browse/DEV-3025)

- [x] Code review
- [ ] Design review
- [x] [This PR Merged](https://github.com/fedspendingtransparency/usaspending-website/pull/1158)

### DIFF
--- a/src/_scss/pages/awardV2/idv/_awardFundingSummary.scss
+++ b/src/_scss/pages/awardV2/idv/_awardFundingSummary.scss
@@ -1,5 +1,6 @@
 .federal-accounts {
   width: 100%;
+  @import 'components/Note';
   .visualization-tooltip {
     @import "components/visualizations/tooltip/_tooltip";
     position: fixed;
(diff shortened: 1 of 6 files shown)


src/_scss/pages/awardV2/idv/_awardFundingSummary.scss
```diff
@@ -1,5 +1,6 @@
 .federal-accounts {
   width: 100%;
+  @import 'components/Note';
```
MaxwellKendall: Not sure I agree with this pattern in our code base... Couldn't we just import all the partials in a parent-partial? It seems to me we're importing the same css multiple times, when we could just import it once. This would appear to remove duplicate CSS.

Webpack removes unused code by default, but I think this is one of the reasons our build takes a long time.

Perhaps I'm wrong about this but I'd like to have a discussion about it!

PR #1168, opened 2019-07-11
Federal Account Components UX

ebdabbs:
**High level description:**

- Adds `Program Activity (TAS)` filter section to the Advanced Search page **in the dev environment only**
- Implements the Federal Account tab and Federal Account Component autocomplete filters (AID and MAIN) with mock data
- Creates a single label with the combined AID and MAIN values once the "Add Filter" button is clicked
- Displays the label in the top filter bar section after a search is submitted

_Note_ - does not include API integration for the autocomplete endpoints or filtering of advanced search results

**Technical details:**

- Makes the shared Autocomplete component more robust by adding:
    - the option to retain values after making a selection
    - a maximum character limit for the input
    - a minimum number of characters required to begin searching
- Adds [Immutable Sets](https://immutable-js.github.io/immutable-js/docs/#/Set) to redux to hold TAS and FA filter values; implements corresponding actions and reducers
- Changes the FontAwesome icon import order to alphabetical 

**JIRA Ticket:**
[DEV-2940](https://federal-spending-transparency.atlassian.net/browse/DEV-2940)

**Mockup:**
https://bahdigital.invisionapp.com/share/6YIAAZS2FUR#/296014843_Advanced_Search_-_TAS_-_Federal_Accounts_Active-Filter_Copy

- [x] Code review
- [x] Design review
- [x] [API contract](https://github.com/fedspendingtransparency/usaspending-api/pull/1874) updated


dlassobah:
hey @ebdabbs I know you note that this is using mock data so I won't spend too much time fussing about the MAIN codes all being 3 characters. however they may be a UX issue here. On other filters that utilize autocomplete (agency, naics, etc), there is a behavior where if the user decides to type out the entire matching phase (re: on agencies you can manually type out "department of agrciulture") but then do not actually select from the dropdown list and instead click out of the field entirely, the behavior is to **clear the field**. The issue we have on TAS is that if the user types out a matching AID for example (020) but then clicks outside the auto-complete, the picklist goes away and it does not allow you to submit your active filter even though you have a valid input. so I would try and match the existing behavior for now

src/js/redux/actions/search/searchFilterActions.js
```diff
@@ -99,6 +99,17 @@ export const updateRecipientLocations = (state) => ({
     location: state
 });
 
+// Program Source Filter
+export const updateFederalAccountComponents = (state) => ({
+    type: 'UPDATE_FEDERAL_ACCOUNT_COMPONENTS',
+    source: state
```
MaxwellKendall: could we rename this param something more descriptive? Just helps w/ readability!

src/js/redux/reducers/search/filters/programSourceFilterFunctions.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * programSourceFilterFunctions.js
+ * Created by Lizzie Salita 6/14/19
+ */
+
+/* eslint-disable import/prefer-default-export */
+// We only have one export but want to maintain consistency with other query modules
+export const updateSelectedSources = (state, sourceIdentifier) => {
```
MaxwellKendall: Same thing can we rename this first param to soemthing a bit more descriptive?

src/js/redux/reducers/search/searchFiltersReducer.js
```diff
@@ -110,6 +117,20 @@ const searchFiltersReducer = (state = initialState, action) => {
             });
         }
 
+        // Program Source (TAS) Filter
+        case 'UPDATE_TREASURY_ACCOUNT_COMPONENTS': {
+            return Object.assign({}, state, {
```
MaxwellKendall: FWIW we can use the spread operator here. I assume you prefer the Object.assign. Makes no difference to me just want to acknowledge this.

we could do:

```javascript
return {
    ...state,
    treasuryAccounts: ProgramSourceFilterFunctions.updateSelectedSources(state.treasuryAccounts, action.source)
}
```

tests/containers/search/topFilterBar/TopFilterBarContainer-test.jsx
```diff
@@ -431,6 +431,31 @@ describe('TopFilterBarContainer', () => {
             expect(filterItem).toEqual(expectedFilterState);
         });
     });
+    it('should update component state with Redux program source filters when available', () => {
+        // mount the container with default props
+        const topBarContainer = setup(defaultProps);
```
MaxwellKendall: Like this `fn` naming `setup`!!!

tests/redux/reducers/search/searchFiltersReducer-test.js
```diff
@@ -28,21 +26,15 @@ describe('searchFiltersReducer', () => {
...
 
         it('should remove a value if currently exists in the set', () => {
             const startingState = Object.assign({}, initialState, {
                 awardType: new Set(['09'])
             });
 
-            expect(
-                searchFiltersReducer(startingState, action).awardType
-            ).toEqual(new Set([]));
+            expect(searchFiltersReducer(startingState, action).awardType).toEqual(new Set([]));
```
MaxwellKendall: much better!!!

MaxwellKendall:
Lots of comments -- Mostly questions and requests to use `.map` instead of `.forEach` and `.filter` instead of `.differenceWith`.

src/js/containers/search/filters/programSource/ProgramSourceContainer.jsx
```diff
@@ -0,0 +1,56 @@
...
+                dirtyFilters={this.dirtyFilters()} />
+        );
+    }
+}
+
+ProgramSourceContainer.propTypes = propTypes;
+
+export default connect(
+    (state) => ({
+        selectedTreasuryComponents: state.filters.treasuryAccounts,
```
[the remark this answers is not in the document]
ebdabbs: These are TAS #s that have been "staged" via `Add Filter` but not yet submitted. Filters in this state are stored in redux, while "pre-staged" filters (for lack of a better term for before `Add Filter`) are stored in the component state.

src/js/containers/search/filters/programSource/ProgramSourceContainer.jsx
```diff
@@ -0,0 +1,56 @@
...
+    selectedTreasuryComponents: PropTypes.object,
+    selectedFederalComponents: PropTypes.object,
+    appliedTreasuryComponents: PropTypes.object,
+    appliedFederalComponents: PropTypes.object,
+    updateFederalAccountComponents: PropTypes.func,
+    updateTreasuryAccountComponents: PropTypes.func
+};
+
+export class ProgramSourceContainer extends React.Component {
+    dirtyFilters() {
```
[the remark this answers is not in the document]
ebdabbs: In this case it means that the staged filters do not match the applied (submitted) filters.

src/js/components/search/filters/programSource/ProgramSourceSection.jsx
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        return (
+            <div className="program-source-filter search-filter">
+                <ul
+                    className="toggle-buttons"
+                    role="menu">
+                    <li>
+                        <button
+                            className={`date-toggle ${activeTreasury}`}
```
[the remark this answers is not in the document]
ebdabbs: This shares styling with the Time Period filter. Same issue in the [location filter](https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/components/search/filters/location/LocationSection.jsx#L85), so we could probably make the class name more generic.

src/js/components/search/filters/programSource/SourceSelectFilter.jsx
```diff
@@ -0,0 +1,166 @@
...
+
+    handleTextInput(programSourceInput) {
+        // Clear existing sources to ensure user can't select an old or existing one
+        this.setState({
+            autocompleteOptions: []
+        });
+
+        // Grab input, clear any exiting timeout
+        const input = programSourceInput.target.value;
+        window.clearTimeout(this.timeout);
```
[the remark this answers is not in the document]
ebdabbs: I discovered via console error that this required using `event.persist()` since synthetic events are pooled - see [this article](https://reactjs.org/docs/events.html#event-pooling)

src/js/components/search/filters/programSource/ProgramSourceSection.jsx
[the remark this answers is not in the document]
ebdabbs: Changed to `tab-toggle`

src/js/redux/actions/search/searchFilterActions.js
ebdabbs: Renamed to `source`

src/js/redux/reducers/search/filters/programSourceFilterFunctions.js
ebdabbs: Renamed to `currentSources`

src/js/components/search/filters/programSource/SourceSelectFilter.jsx
```diff
@@ -0,0 +1,166 @@
...
+        if (input.length >= 1) {
+            this.setState({
+                searchString: input
+            });
+
+            let autocompleteOptions = [];
+
+            const matches = this.props.options.filter((source) => (source.code.includes(this.state.searchString)));
+            // Filter out any selectedSources that may be in the result set
+            if (this.props.selectedSources && this.props.selectedSources.length > 0) {
```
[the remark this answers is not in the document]
ebdabbs: I realized we only need the first filter. We don't need to prevent users from picking a value that's already been selected, because the component (AID for example) is part of the larger Federal Account number and could repeat in different filters (e.g. `020-1234`, `020-5678`).

ebdabbs:
@MaxwellKendall Good questions & comments. I think many of these instances are code copied from other filters written before we standardized these patterns (`map`, `throttle`/`debounce`), and you're right that we should take advantage of them as we add new features.

ebdabbs:
@dlassobah After testing this out I realized an issue with clearing the selection when a user clicks outside the autocomplete component. It will clear when you go to type in the next filter or to click `Add Filter`. That functionality works for other Autocompletes because your selection is immediately staged when you select from the dropdown. The City filter has more similar functionality to this one, and does not clear the input when you click outside. Might be something to consider when we look into standardizing the various autocomplete/dropdown elements!

src/js/components/sharedComponents/autocomplete/Autocomplete.jsx
```diff
@@ -79,7 +95,9 @@ export default class Autocomplete extends React.Component {
 
         target.addEventListener('blur', () => {
```
MaxwellKendall: This `fn` setUpAutocomplete should be decomposed into two functions `handleBlur` and `handleOnKeyDown`. 

We also should avoid using `addEventListener` in favor of JSX event attributes wherever possible.

I defer to your judgment as to whether to address this tech debt in your PR!

ebdabbs: Planning to address this as part of the Autocomplete refactor for [DEV-2921](https://federal-spending-transparency.atlassian.net/browse/DEV-2921)

PR #1170, opened 2019-07-11
Add Congressional District to Recipient Address

ebdabbs:
**High level description:**

Recipient Details Panel
- Adds Congressional District to the Recipient Address
- Displays `--` when no location information is available
- Displays `--` instead of `Unknown` in the panel when both recipient name and hash are null

Recipient Overview section
- Displays an unlinked name when recipient hash is null
- Displays `Unknown` when the recipient name is null

**JIRA Ticket:**
Follow up to [DEV-2740](https://federal-spending-transparency.atlassian.net/browse/DEV-2740) (Failed Test)

**Mockup:**
https://bahdigital.invisionapp.com/share/GVIAAHOBER9#/296001079_Award_Summary_2-0_-_IDV_-future-

- [x] Code review

### DIFF
--- a/src/js/components/awardv2/shared/additionalInfo/RecipientDetails.jsx
+++ b/src/js/components/awardv2/shared/additionalInfo/RecipientDetails.jsx
@@ -102,6 +102,7 @@ export default class RecipientDetails extends React.Component {
                                 <td>
                                     <div>{this.props.data.location.streetAddress}</div>
                                     <div>{this.props.data.location.regionalAddress}</div>
+                                    <div>{this.props.data.location.fullCongressionalDistrict}</div>
                                 </td>
                             </tr>
                             <tr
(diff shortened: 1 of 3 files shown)


src/js/components/awardv2/shared/additionalInfo/RecipientDetails.jsx
```diff
@@ -76,7 +90,7 @@ export default class RecipientDetails extends React.Component {
                                 className="accordion-table__row">
                                 <td>Recipient</td>
                                 <td>
-                                    {this.formatRecipientLink(this.props.data.internalId, this.props.data.name)}
+                                    {this.formatRecipientLink(this.props.data.internalId, this.props.data._name)}
```
jonhill13: Should this be `name`? without the underscore?

ebdabbs: Good question - `name` maps to `Unknown` for nulls, whereas `_name` maps to the empty string for nulls. At the top of the page in the recipient section we use `name` to display Unknown, but here I'm using the falsy-ness of empty string to show `--` instead.

jonhill13: wont is just display an empty string since `_name` will be an empty string if it is `null` and `formatRecipientLink` will just display the param `name` if there is no `internalId` and no `name`?

jonhill13: nevermind, i thought formatRecipientLink was the same as Agency Recipient

PR #1184, opened 2019-07-25
TAS component filters API integration

ebdabbs:
**High level description:**

Connects the TAS component filters to their respective autocomplete endpoints.

**Technical details:**

- Creates an autocomplete container for program source component filters that selects the correct helper function to make the API call, and uses selected filters as parameters to narrow down the results
- Moves information about the component filters to a data mapping file
- Implements the helper functions for API calls to each of the 7 autocomplete component endpoints
- Fixes UX bugs related to clearing invalid values and displaying error messages correctly
- Adds unit tests for the new container

**JIRA Ticket:**
[DEV-2921](https://federal-spending-transparency.atlassian.net/browse/DEV-2921)

**Mockup:**
https://bahdigital.invisionapp.com/share/8SIAAYD25TJ#/296003145_Advanced_Search_-_TAS_-_Federal_Accounts

- [x] Code review
- [x] Design review @dlassobah 
- [x] API [#1917](https://github.com/fedspendingtransparency/usaspending-api/pull/1917) merged


dlassobah:
@ebdabbs I'm still seeing some differences in how these are input vs other auto-complete filters such as agency. For 'agency', you type and narrow down your results and you MUST click on one of the suggestions provided. if you do not, it actually clears your input alltogether. In the case of TAS, it does not clear- and to make things even more potentially confusing, if I manually type 014 into AID for example, which is a valid entry, but do not click on the auto-completed suggestion, it will remain in the text field without allowing me to submit the search.

And one last future consideration as you begin to develop the treasury account tab components- did Ross give us go ahead to make ALL components auto-complete? if he did, we would likely want to remove the 'add filter' button alltogether in order to again align to those existing patterns. We should be focused on getting up something functional at this point however so we can revisit that change after MVP due to all the variables still flying around this one.

ebdabbs:
@dlassobah check out this PR: #1185 for the TAS tab. I made some updates to the filters so that it will clear the text box if you don't select a value from the dropdown. 
Yes, we did decide to make all of the components autocomplete.

dlassobah:
UX issues resolved

PR #1193, opened 2019-08-02
TAS Autocomplete Bug Fix

ebdabbs:
**High level description:**
Fixes a bug that caused TAS Component autocomplete selections to get staged after being deleted from the text input.

**Technical details:**
Previously, `handleInputText` was only getting called if the input value was truthy, which prevented `clearSelection` from getting called when the entire value was deleted.

**JIRA Ticket:**
[DEV-3219](https://federal-spending-transparency.atlassian.net/browse/DEV-3219)

**Mockup:**
N/A

- [x] Code review

### DIFF
--- a/src/js/components/sharedComponents/autocomplete/Autocomplete.jsx
+++ b/src/js/components/sharedComponents/autocomplete/Autocomplete.jsx
@@ -83,10 +83,8 @@ export default class Autocomplete extends React.Component {
         e.persist();
         this.checkValidity(e.target.value);
         let selectedIndex = 0;
-        if (e.target.value) {
-            this.props.handleTextInput(e);
-        }
-        else {
+        this.props.handleTextInput(e);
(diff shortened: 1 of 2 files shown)


MaxwellKendall:
@ebdabbs 

Was still able to duplicate this error locally. Here are my steps to reproduce:

1. Select ATA, AID & MAIN
2. Highlight ATA selection and delete

**Expected Results:** Add filter is now disabled
**Actual Results:** Add filter is still enabled

ebdabbs:
@MaxwellKendall I would still expect Add Filter to be enabled in that case, since AID and MAIN are the only two required values

MaxwellKendall:
@ebdabbs my bad misunderstood the requirements!! Was thinking ATA was a required field!

Local testing passed!

MaxwellKendall:
Also confirmed locally in various browsers!

PR #1194, opened 2019-08-05
DEV-1987 

emily-brents:
**High level description:**

Added Direct Payments, Loans, and Other type codes to the allowed types for the subgrants tab.

**Technical details:**
Concatenated `awardTypeGroups` in `awardType.js` to allow the subgrants tab to send the codes from Direct Payments, Loans, as well as grants.

**JIRA Ticket:**
[DEV-1987](https://federal-spending-transparency.atlassian.net/browse/DEV-1987)

- [x] Code review
- [N/A] Design review (if applicable)
- [N/A] [API contract](https://github.com/fedspendingtransparency/usaspending-api/tree/dev/usaspending_api/api_contracts) updated (if applicable)
- [x] [API #1953](https://github.com/fedspendingtransparency/usaspending-api/pull/1953) merged (if applicable)
- [N/A] Verified cross-browser compatibility

### DIFF
--- a/src/js/dataMapping/search/awardType.js
+++ b/src/js/dataMapping/search/awardType.js
@@ -88,5 +88,5 @@ export const awardTypeGroupLabels = {
 
 export const subawardTypeGroups = {
     subcontracts: awardTypeGroups.contracts.concat(awardTypeGroups.idvs),
-    subgrants: awardTypeGroups.grants
+    subgrants: awardTypeGroups.grants.concat(awardTypeGroups.direct_payments).concat(awardTypeGroups.loans).concat(awardTypeGroups.other)
 };


kbard:
So it doesn't explicitly state this in the ticket, but this filter doesn't seem to be affecting downloads at all.  Should it be?

emily-brents:
> So it doesn't explicitly state this in the ticket, but this filter doesn't seem to be affecting downloads at all. Should it be?

yeah, I assume so.

kbard:
> So it doesn't explicitly state this in the ticket, but this filter doesn't seem to be affecting downloads at all. Should it be?

This isn't an issue.  It appears to be working as expected.  My bad.

PR #1197, opened 2019-08-07
TAS search tooltips

ebdabbs:
**High level description:**
Adds info tooltips to each of the Program Source (TAS) filter components. 

**Technical details:**
Uses a CSS-activated tooltip, which is compatible with collapsible sections. 

**JIRA Ticket:**
[DEV-2824](https://federal-spending-transparency.atlassian.net/browse/DEV-2824)

**Mockup:**
https://bahdigital.invisionapp.com/share/Z7IAB1I53Y4#/296002937_Advanced_Search_-_TAS (click the `i` icon next to ATA)

- [x] Code review
- [x] Design review @dlassobah 

### DIFF
--- a/src/_scss/pages/search/filters/programSource/programSource.scss
+++ b/src/_scss/pages/search/filters/programSource/programSource.scss
@@ -37,6 +37,41 @@
             .program-source-select-filter__label {
                 @include display(flex);
                 @include justify-content(space-between);
+                @import '../../../../components/visualizations/tooltip/csstooltip';
+                .tooltip-popover-container {
+                    margin-left: rem(5);
+                    color: $color-gray-lighter;
+                    &:hover, &:focus {
+                        color: $color-gray-medium;
+                    }
+                    .tooltip-popover {
+                        // override the positioning to be centered over the icon
+                        bottom: rem(35);
+                        left: rem(-145);
+                        padding: 0;
+                        font-size: $small-font-size;
+                        &.program-source-info-tooltip {
+                            box-shadow: $box-shadow;
+                            background-color: $color-white;
+                            border: solid rem(1) $color-gray-light;
+                            .program-source-info-tooltip__heading {
+                                background-color: $color-gray-lightest;
+                                border-top-left-radius: rem(5);
+                                border-top-right-radius: rem(5);
+                                padding: rem(10) rem(15);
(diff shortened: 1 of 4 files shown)


dlassobah:
@ebdabbs functions as expected. I know these definitions were all pulled from the RSS data source, but the definition for SUB is a little odd to me. It starts with 'this is a component of TAS' which seems unnecessary to me. Maybe it's something we should run by @ross-williford to see if we want to make an exception on that definition.

src/_scss/pages/search/filters/programSource/programSource.scss
```diff
@@ -36,8 +36,47 @@
             @import "elements/filters/_typeahead";
             .program-source-select-filter__label {
                 @include display(flex);
-                @include justify-content(space-between);
+                @import '../../../../components/visualizations/tooltip/csstooltip';
```
MaxwellKendall: @ebdabbs,

not in scope for this pr but would like to discuss the performance effect of @import-ing scss like this multiple times in multiple places.

This @import is only used once, but there are many places we're using this pattern and importing the same piece of code multiple times which I'm afraid may be bloating our compiled css.

src/_scss/pages/search/filters/programSource/programSource.scss
```diff
@@ -36,8 +36,47 @@
...
+                    @include display(flex);
+                    @include flex (0 0 auto);
+                    margin-left: rem(5);
+                    color: $color-gray-lighter;
+                    &:hover, &:focus {
+                        color: $color-gray-medium;
+                    }
+                    .tooltip-popover {
+                        // override the positioning to be centered over the icon
+                        bottom: rem(35);
```
MaxwellKendall: Looks like this value isn't always reliable with the first and third icons. These are the icons which have to the left of them a <label> component that overflows/wraps to the second line like so:

Allocation Transfer Agency Identifier
(ATA)
Beginning Period of Availability
(BPOA)

I would guess this is the root cause as all the other icons work perfectly and the <label> to the left does not wrap.

ebdabbs: Good catch. I changed the icon alignment so that it's not totally obscured, since dynamically measuring seems overkill.

PR #1199, opened 2019-08-08
[DEV-3004] Implementing new Environment Variable to Determine API URL

MaxwellKendall:
**High level description:**
Instead of using a secret `globalConstant` file, we're using a public one with the `API` property defined by an environment variable.

**Technical details:**
Using webpack to expose environment variables for global configuration. Also sets env-variables to defaults if they're not defined.

**JIRA Ticket:**
[DEV-3004](https://federal-spending-transparency.atlassian.net/browse/DEV-3004)

- [ ] Code review
- [ ] Test in sandbox once [this pr is merged](https://github.com/fedspendingtransparency/usaspending-config/pull/518)
- [ ] Verified cross-browser compatibility


webpack/webpack.common.js
```diff
@@ -68,20 +66,6 @@ module.exports = {
...
-                    {
-                        loader: "sass-loader",
-                        options: {
-                            sourceMap: !isProduction,
-                            includePaths: ["./src/_scss", "./node_modules"]
-                        }
-                    }
-                ]
-            },
             {
```
MaxwellKendall: `isProduction` was always false here because `process.env.NODE_ENV` was always undefined.

PR #1204, opened 2019-08-12
[DEV-3004] Hiding bash-script, removing sample globalConstants, env var API --> USASPENDING_API

MaxwellKendall:
**High level description:**
Follow up remove sampleGlobalConstants and hide bash script to help team members with local dev

**Technical details:**
Adds to gitiginore

**JIRA Ticket:**
[DEV-3004](https://federal-spending-transparency.atlassian.net/browse/DEV-3004)

- [x] Code review


src/js/GlobalConstants.js
```diff
@@ -4,7 +4,7 @@
  */
 
 const globalConstants = {
-    API: process.env.API,
+    API: process.env.USASPENDING_API,
```
ebdabbs: Should this be called `USASPENDING_API` to match dev.config and prod.config?

MaxwellKendall: The only place we're directly using `process.env.USASPENDING_API` is here. If we change the `globalConstants.API` to `globalConstants.USASPENDING_API`we'd have to update like 100 files.

PR #1208, opened 2019-08-15
DEV-3184: Allow activity tooltip to handle missing recipient hash

kbard:
**High level description:**
IDV activity chart was not handling missing/null recipient hash (recipient_id).  Now, if the hash (id) is missing, the name will display (or `--` if the name is also missing).

**JIRA Ticket:**
[DEV-3184](https://federal-spending-transparency.atlassian.net/browse/DEV-3184)

- [x] Code review

### DIFF
--- a/src/js/components/awardv2/idv/activity/ActivityChartTooltip.jsx
+++ b/src/js/components/awardv2/idv/activity/ActivityChartTooltip.jsx
@@ -177,7 +177,7 @@ export default class IdvActivityTooltip extends React.Component {
             (
                 <div>
                     This IDV &#62; {this.getLinks(
-                        `award`,
+                        'award',
                         data.parentGeneratedId,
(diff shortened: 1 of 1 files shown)


src/js/components/awardv2/idv/activity/ActivityChartTooltip.jsx
```diff
@@ -178,7 +177,8 @@ export default class IdvActivityTooltip extends React.Component {
             (
                 <div>
                     This IDV &#62; {this.getLinks(
-                        `award/${data.parentGeneratedId}`,
+                        `award`,
```
jonhill13: can we change this to single quotes instead of backticks since we do not have any data in those strings anymore.  The back ticks are template literals in JS and require an extra step for babel to read and translate them down to regular JS.

PR #1259, opened 2019-09-18
Dev-3095 V2 Award Pages Dates Section

jonhill13:
**High level description:**

Award V2 Pages now have a Dates Section

**Technical details:**

remove AwardDates and replace it with IdvDates

**JIRA Ticket:**
[DEV-3095](https://federal-spending-transparency.atlassian.net/browse/DEV-3095)

**Mockup:**
See ticket

- [x] Code review
- [x] Design review (@AGuyNamedMarco )
- [x] All `componentWillReceiveProps`, `componentWillMount`, and `componentWillUpdate` in relevant child/parent components/containers replaced with [future compatible life-cycle methods](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) per [JIRA item 3341](https://federal-spending-transparency.atlassian.net/browse/DEV-3341)


jonhill13:
@MaxwellKendall it does show try http://localhost:3000/#/award_v2/65207739

MaxwellKendall:
@jonhill13 

For some reason this doesn't look right for this award: https://dev.usaspending.gov/#/award_v2/67508500

Staging also shows the today line closer to the middle and 3 years remaining whereas locally it shows the today line close to the end, even though three years are left.

Locally it also doesn't show any text about how much time is remaining:

**DEV/STAGING:**
image

**LOCAL:**
image

MaxwellKendall:
@jonhill13 

(Looks like previously)[https://github.com/fedspendingtransparency/usaspending-website/pull/1259/files?file-filters%5B%5D=.jsx&hide-deleted-files=true#diff-87e79c0b6d823eefcaa3ef8089a9fa9fL20] we were comparing the `startDate` with the _potential_ end date.

This award, with your changes, is comparing the `startDate` with the `endDate` or more precisely -- _current end date_ -- which is `9/30/19`, instead of the _potential end date_ which is `09/30/22`

Looks like we should be comparing the startDate to the potential end date to keep things the same!

src/js/models/v2/awardsV2/BaseFinancialAssistance.js
```diff
@@ -54,7 +54,8 @@ BaseFinancialAssistance.populate = function populate(data) {
     if (data.period_of_performance) {
         const periodOfPerformanceData = {
             startDate: data.period_of_performance.start_date,
-            endDate: data.period_of_performance.end_date
+            endDate: data.period_of_performance.end_date,
+            lastModifiedDate: data.period_of_performance.last_modified_date
         };
```
MaxwellKendall: I feel like this should be in the CoreAward Model? Would like to hear your justification for putting it here before approving.

jonhill13: I put it here because that is where it is for contracts and that is where the other dates are housed.

MaxwellKendall: NVM looking at the code I see what is going on.

PR #1356, opened 2019-10-30
Dev-3814 Contract Related Awards Section

jonhill13:
**High level description:**

Adds the related awards section to the contracts v2 page.

**Technical details:**

* add counts from contract subawards to related awards section

* add subawards section for contracts

* dd another jumptosection for subawards tab in history table

* update post text

**JIRA Ticket:**
[DEV-3814](https://federal-spending-transparency.atlassian.net/browse/DEV-3814)

**Mockup:**
https://bahdigital.invisionapp.com/share/9EIABE5KQ24#/screens/296011639

The following are ALL required for the PR to be merged (in ascending LOE):
- [x] Link to this PR in JIRA ticket
- [x] Tagged Design, Testing, and Code Reviewers in `#us-ux-frontend` Slack Channel in the thread under the Post from Github for their SA
- [ ] Scheduled demo including Design/Testing/Front-end `if applicable`
- [N/A] Provided instructions for testing in JIRA (for benefit of testing) and PR (for benefit of code reviewer) `if applicable`
- [N/A] Added Unit Tests for methods in Container Components, reducers, and helper functions `if applicable`
- [x] All componentWillReceiveProps, componentWillMount, and componentWillUpdate in relevant child/parent components/containers replaced with [future compatible life-cycle methods](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) per [JIRA item 3339](https://federal-spending-transparency.atlassian.net/browse/DEV-3339)
- [N/A] [API contract](https://github.com/fedspendingtransparency/usaspending-api/tree/dev/usaspending_api/api_contracts) updated `if applicable`
- [N/A] [API #1234](https://github.com/fedspendingtransparency/usaspending-api/pull/1234) merged `if applicable`
- [ ] Code review complete

### DIFF
--- a/src/js/components/awardv2/shared/overview/RelatedAwards.jsx
+++ b/src/js/components/awardv2/shared/overview/RelatedAwards.jsx
@@ -15,6 +15,7 @@ const propTypes = {
     overview: PropTypes.object,
     jumpToSection: PropTypes.func,
     setRelatedAwardsTab: PropTypes.func,
+    jumpToSubAwardHistoryTable: PropTypes.func,
     counts: PropTypes.object
 };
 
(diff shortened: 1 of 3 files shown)


src/js/components/awardv2/shared/overview/RelatedAwards.jsx
```diff
@@ -42,47 +43,67 @@ export default class RelatedAwards extends React.Component {
...
-                count: formatNumber(counts.grandchild_awards),
-                name: 'Grandchild Award',
-                funcName: 'GrandchildAwards',
-                glossary: 'award'
-            }
-        ];
-
+        let childData = [];
+        if (overview.category === 'idv') {
+            childData = [
```
MaxwellKendall: Ideally, this would be in a datamapping file, where we would import it and map over to add `count` and `postText` properties. Not a huge deal we're doing it this way though due to the use of dynamic values for these properties.

Defer to your judgment / preference as to whether we do this.

src/js/components/awardv2/shared/overview/RelatedAwards.jsx
```diff
@@ -42,47 +43,67 @@ export default class RelatedAwards extends React.Component {
...
+        }
         return (
             <div className="related-awards__label related-awards__label_count">
                 <div className="related-awards__counts">
                     {map(childData, (data) => (
                         <button
                             key={`${data.glossary}count`}
                             className="award-viz__button"
-                            onClick={this[`jumpToReferencedAwardsTable${data.funcName}Tab`]}>
+                            onClick={this[`${data.funcName}`]}>
```
MaxwellKendall: This is an anti-pattern because if we wanted to see all the callers of `jumpToAwardHistoryTableSubAwardsTab` or one of the other `childData[i].funcName`, we couldn't rely on a basic global cntrl + f search.

One solution is to use a map inside the render where we map `funcName` to the class method. So then, we can see where the map is used to identify the caller.

In this context though, I don't think it's a big deal. I'm sure we're doing this elsewhere and it's just not a good pattern to use uncritically.

MaxwellKendall: Again, defer to your judgment as to whether or not to clean this up now.

PR #1442, opened 2019-12-20
[data-transparency-ui] migrating to semver for data-transparency-ui dependency

MaxwellKendall:
Corresponding documentation [here](https://github.com/fedspendingtransparency/data-transparency-ui/pull/20)

- [x] local test
- [x] sandbox test

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4764,8 +4764,8 @@
             }
         },
         "data-transparency-ui": {
-            "version": "github:fedspendingtransparency/data-transparency-ui#9d7e905c0eed442a068c65f5f0f3870876719174",
-            "from": "github:fedspendingtransparency/data-transparency-ui#master",
+            "version": "github:fedspendingtransparency/data-transparency-ui#402022054ec84c30a87df05dc00a4e9cdd07d475",
+            "from": "github:fedspendingtransparency/data-transparency-ui#v0.1.0",
             "requires": {
                 "@fortawesome/fontawesome-svg-core": "^1.2.25",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -4764,8 +4764,8 @@
             }
         },
         "data-transparency-ui": {
-            "version": "github:fedspendingtransparency/data-transparency-ui#9d7e905c0eed442a068c65f5f0f3870876719174",
-            "from": "github:fedspendingtransparency/data-transparency-ui#master",
+            "version": "github:fedspendingtransparency/data-transparency-ui#402022054ec84c30a87df05dc00a4e9cdd07d475",
```
kwhickey: Is this something NPM does? -- turn the tag into a git commit hash?
Just curious if that tag is updated to point to a different hash, will this lock file regen automatically, or is it up to the developer that generated the lock file to know to re-generate it now that the tag changed?

MaxwellKendall: @kwhickey,

Yes. 

IMO, It ought to be our recommended process that when updating the data-transparency-ui semver number, the dev should do so indirectly by running:

`npm uninstall data-transparency-ui && npm install fedspendingtransparency/data-transparency-ui#vX.X.X`

This way, the package.json will be updated and the corresponding lock file will be updated, only for the data-transparency-package.

MaxwellKendall: Documentation provided in the UI library [here](https://github.com/fedspendingtransparency/data-transparency-ui/pull/20)

PR #1514, opened 2020-02-13
[DEV-4504] decoding award IDs on POST request Objects, adding tests for every case

MaxwellKendall:
## Do not merge until we test in sandbox

**High level description:**
In the context of a post request, we should be `decoding` the generated_unique_award_ids that are encoded in our models.

**Technical details:**
API expects raw/non-ecoded award ids in post requests.
Also added a bunch of tests to document how we are supposed to handle these values in every context.

**JIRA Ticket:**
[DEV-4504](https://federal-spending-transparency.atlassian.net/browse/DEV-4504)

Author:
- [x] Added Unit Tests for methods in Container Components, reducers, and helper functions `if applicable`
- [x] Tested in sandbox

Reviewer(s):
- [ ] Code review complete

### DIFF
--- a/src/js/containers/keyword/table/ResultsTableContainer.jsx
+++ b/src/js/containers/keyword/table/ResultsTableContainer.jsx
@@ -192,7 +192,7 @@ export default class ResultsTableContainer extends React.Component {
         };
 
         this.searchRequest = KeywordHelper.performKeywordSearch(params);
-        this.searchRequest.promise
+        return this.searchRequest.promise
             .then((res) => {
                 const newState = {
                     inFlight: false
--- a/src/js/containers/search/table/ResultsTableContainer.jsx
+++ b/src/js/containers/search/table/ResultsTableContainer.jsx
@@ -336,7 +336,7 @@ export class ResultsTableContainer extends React.Component {
 
         // Set the params needed for download API call
         this.searchRequest = SearchHelper.performSpendingByAwardSearch(params);
-        this.searchRequest.promise
+        return this.searchRequest.promise
             .then((res) => {
                 const newState = {
                     inFlight: false
--- a/src/js/helpers/apiRequest.js
+++ b/src/js/helpers/apiRequest.js
@@ -21,10 +21,20 @@ export const apiRequest = (axiosParams = {}) => {
     };
 
     const params = (options = {}) => {
-        const parameters = { ...defaultParams };
-        Object.assign(parameters, options);
-        parameters.headers = headers(options.headers);
-        return parameters;
+        return Object
+            .keys(options)
+            .reduce((acc, key) => {
+                if (key === 'data' && Object.keys(options.data).includes('award_id')) {
+                    return {
+                        ...acc,
+                        data: {
+                            ...options.data,
+                            award_id: decodeURIComponent(options.data.award_id)
+                        }
+                    };
+                }
+                return { ...acc, [key]: options[key] };
+            }, { ...defaultParams, headers: headers(options.headers) });
     };
 
     const cancel = () => cancelToken.cancel();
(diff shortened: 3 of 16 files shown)


src/js/containers/keyword/table/ResultsTableContainer.jsx
```diff
@@ -192,7 +192,7 @@ export default class ResultsTableContainer extends React.Component {
         };
 
         this.searchRequest = KeywordHelper.performKeywordSearch(params);
-        this.searchRequest.promise
+        return this.searchRequest.promise
```
MaxwellKendall: allows us to test the class method where this is invoked.

src/js/containers/search/table/ResultsTableContainer.jsx
```diff
@@ -336,7 +336,7 @@ export class ResultsTableContainer extends React.Component {
 
         // Set the params needed for download API call
         this.searchRequest = SearchHelper.performSpendingByAwardSearch(params);
-        this.searchRequest.promise
+        return this.searchRequest.promise
```
MaxwellKendall: allows us to test the class method where this is invoked.

src/js/helpers/apiRequest.js
```diff
@@ -21,10 +21,20 @@ export const apiRequest = (axiosParams = {}) => {
...
-        return parameters;
+        return Object
+            .keys(options)
+            .reduce((acc, key) => {
+                if (key === 'data' && Object.keys(options.data).includes('award_id')) {
+                    return {
+                        ...acc,
+                        data: {
+                            ...options.data,
+                            award_id: decodeURIComponent(options.data.award_id)
```
MaxwellKendall: This value needs to be raw/un encoded

src/js/helpers/apiRequest.js
```diff
@@ -20,12 +20,20 @@ export const apiRequest = (axiosParams = {}) => {
         return defaultHeaders;
     };
 
-    const params = (options = {}) => {
-        const parameters = { ...defaultParams };
-        Object.assign(parameters, options);
-        parameters.headers = headers(options.headers);
-        return parameters;
-    };
+    const params = (options = {}) => Object
```
jonhill13: What made you want to use reduce here? Couldn't we just `if (params.data && params.award_id) params.data.award_id = decodeURIComponent(options.data.award_id);`

src/js/helpers/apiRequest.js
```diff
@@ -20,12 +20,20 @@ export const apiRequest = (axiosParams = {}) => {
...
+                return {
+                    ...acc,
+                    data: {
+                        ...options.data,
+                        award_id: decodeURIComponent(options.data.award_id)
+                    }
+                };
+            }
+            return { ...acc, [key]: options[key] };
+        }, { ...defaultParams, headers: headers(options.headers) });
```
jonhill13: Personally, I would add a `post` method, considering we would want to check all our posts if there is an `award_id`.

Something like:
```
const post = (parameters) => {
  const newParams = params(parameters);
  /**
  * Removes the encoding of the award_id.
  */
  if (newParams.data && newParams?.data.award_id) {
    // maybe check here to make sure it is encoded?
    newParams.data.award_id = decodeURIComponent(options.data.award_id);
  }
  Axios.request(newParams);
};
```
This way we will have isolated new defaults for our posts, and in the future we can do so with our gets, puts, etc...

Although, we would need to update the post routes to be posts. But since post is not default any post we have should be searchable by `method: 'post'`.  

I originally was going to set it up this way but there was no real need for it.

src/js/helpers/apiRequest.js
MaxwellKendall: You're right it just makes it more confusing to use reduce.

MaxwellKendall:
@jonhill13 since this is a hotfix resolving a bug in production can we merge as is and move this refactoring outside the scope of this delivery?

PR #1642, opened 2020-04-30
[DEV-4278] Optimizing getTasNodeFromTree for Performance

MaxwellKendall:
**High level description:**
This `fn` takes an array of hierarchical nodes and a code and returns the node with the given code, from the given tree.

**Technical details:**
This `fn` was using `.forEach` to do this, which means it was continuing to iterate through the current node hierarchy level even after it found the correct node.

Now, we're using `.find` to get the node at the current level, then using `.reduce` to recursively look at the next level. The benefit of this is that `.find` breaks the iteration once we get the node. `.reduce`, though, I'm assuming has the same problem as `.forEach`.

Feedback on how to optimize this further is welcome and needed! Perhaps we could just use `.find` recursively or optimize this a bit better.

**JIRA Ticket:**
[DEV-4278](https://federal-spending-transparency.atlassian.net/browse/DEV-4278)

Author:
`N/A` Added Unit Tests for methods in Container Components, reducers, and helper functions `if applicable`

Reviewer(s):
- [ ] Code review complete

### DIFF
--- a/src/js/helpers/tasHelper.js
+++ b/src/js/helpers/tasHelper.js
@@ -37,39 +37,18 @@ export const cleanTasData = (nodes) => cleanTreeData(nodes, tasKeyMap)
         return node;
     });
 
-export const getTasNodeFromTree = (tree, id) => {
-    let selectedNode;
-    tree
-        .forEach((agency) => {
-            if (agency.value === id) selectedNode = agency;
+export const getTasNodeFromTree = (tree, code) => {
+    if (!tree) return tree;
+    const topLevelNode = tree.find((node) => node.value === code);
+    if (topLevelNode) return topLevelNode;
+    return tree
+        .reduce((acc, node) => {
+            if (acc) return acc;
+            if (node.value === code) return node;
+            return getTasNodeFromTree(node.children, code);
         });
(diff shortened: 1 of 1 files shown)


src/js/helpers/tasHelper.js
```diff
@@ -37,39 +37,18 @@ export const cleanTasData = (nodes) => cleanTreeData(nodes, tasKeyMap)
...
-                            federalAccount.children
-                                .forEach((tas) => {
-                                    if (tas.value === id) selectedNode = tas;
-                                });
-                        }
-                    });
-            }
-        });
-    return selectedNode;
+export const getTasNodeFromTree = (tree, code) => {
```
jonhill13, line 40: findTasNode

src/js/helpers/tasHelper.js
```diff
@@ -37,39 +37,18 @@ export const cleanTasData = (nodes) => cleanTreeData(nodes, tasKeyMap)
...
+    if (!tree) return tree;
```
jonhill13, line 41: to me, this is confusing, and reason we return tree? i guess your returning null? i would rather see return null

PR #1652, opened 2020-05-04
Y-Domain Bug Contract Grant Activity

jonhill13:
**High level description:**

> • update setting the y domain to a cleaner solution

- [x] Code review complete

### DIFF
--- a/src/js/components/award/shared/activity/ContractGrantActivityChart.jsx
+++ b/src/js/components/award/shared/activity/ContractGrantActivityChart.jsx
@@ -1,6 +1,6 @@
 import React, { useState, useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
-import { cloneDeep, compact, sum, maxBy } from 'lodash';
+import { cloneDeep, compact, sum } from 'lodash';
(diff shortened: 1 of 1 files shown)


src/js/components/award/shared/activity/ContractGrantActivityChart.jsx
```diff
@@ -87,15 +87,19 @@ const ContractGrantsActivityChart = ({
...
-        const yZero = clonedTransactions.length > 1 ?
-            clonedTransactions[0].running_obligation_total :
-            0;
-        let yOne = !clonedTransactions.length ?
-            0 :
-            totalObligation || clonedTransactions.pop().running_obligation_total;
-        // if any transaction is greater than the obligation
-        const transactionIsGreaterThanObligation = clonedTransactions.find((t) => t.running_obligation_total > totalObligation);
-        if (transactionIsGreaterThanObligation) yOne = transactionIsGreaterThanObligation.running_obligation_total;
+        let yZero = 0;
```
ebdabbs, line 90: Looks like this can be a `const` since it gets set to zero in both cases.

PR #1656, opened 2020-05-05
[DEV-4528] Adding Recursion and a few Helpers to Clean Up Logic

MaxwellKendall:
**High level description:**
Fns used to incrementally populate the tree were not optimized to handle the four levels of PSC.

**Technical details:**
When performing a search, we populate the tree partially w/ the search results and placeholders. Then, when a user leaves search and expands down the tree, we complete that partial population. In order to preserve the proper checked state (full or half checked) we have to ensure that as we complete the tree population, we do not overwrite existing nodes from search.

We were doing this in PSC because we weren't anticipating 3 levels of children. **Now, we are via recursion**.

There is some added clean up to the logic as well by extracting some new helper fns outside the scope of the larger complex functions.

**JIRA Ticket:**
[DEV-4528](https://federal-spending-transparency.atlassian.net/browse/DEV-4528)

Author:
- [x] Added Unit Tests for methods in Container Components, reducers, and helper functions `if applicable`

Reviewer(s):
- [x] Code review complete

### DIFF
--- a/src/js/containers/search/filters/naics/NAICSContainer.jsx
+++ b/src/js/containers/search/filters/naics/NAICSContainer.jsx
@@ -197,19 +197,24 @@ export class NAICSContainer extends React.Component {
     }
 
     onUncheck = (newChecked, uncheckedNode) => {
-        const [stagedNaicsFilters, newUnchecked] = decrementNaicsCountAndUpdateUnchecked(
-            uncheckedNode,
-            this.props.unchecked,
-            this.props.checked,
-            this.state.stagedNaicsFilters,
-            this.props.nodes
-        );
+        if (uncheckedNode.checked) {
+            this.onCheck(newChecked);
+        }
+        else {
+            const [stagedNaicsFilters, newUnchecked] = decrementNaicsCountAndUpdateUnchecked(
+                uncheckedNode,
+                this.props.unchecked,
+                this.props.checked,
+                this.state.stagedNaicsFilters,
+                this.props.nodes
+            );
 
-        this.props.setUncheckedNaics(newUnchecked);
-        this.props.stageNaics(newChecked, newUnchecked, stagedNaicsFilters);
-        this.props.setCheckedNaics(newChecked);
+            this.props.setUncheckedNaics(newUnchecked);
+            this.props.stageNaics(newChecked, newUnchecked, stagedNaicsFilters);
+            this.props.setCheckedNaics(newChecked);
 
-        this.setState({ stagedNaicsFilters });
+            this.setState({ stagedNaicsFilters });
+        }
     }
 
     onExpand = (value, expanded, fetch) => {
--- a/src/js/containers/search/filters/psc/PSCCheckboxTreeContainer.jsx
+++ b/src/js/containers/search/filters/psc/PSCCheckboxTreeContainer.jsx
@@ -15,7 +15,10 @@ import {
     getPscNodeFromTree
 } from 'helpers/pscHelper';
 import { fetchPsc } from 'helpers/searchHelper';
-import { removePlaceholderString } from 'helpers/checkboxTreeHelper';
+import {
+    getAllDescendants,
+    removePlaceholderString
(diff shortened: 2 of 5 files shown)


src/js/containers/search/filters/psc/PSCCheckboxTreeContainer.jsx
```diff
@@ -172,7 +175,7 @@ export class PSCCheckboxTreeContainer extends React.Component {
                 const node = getPscNodeFromTree(nodes, expandedAndChecked);
                 return [
                     ...acc,
-                    ...expandPscNodeAndAllDescendantParents([node])
+                    ...getAllDescendants(node)
```
MaxwellKendall, line 178: We needed the leaf level nodes to properly update the checked array here.

src/js/helpers/checkboxTreeHelper.js
```diff
@@ -57,7 +57,7 @@ export const removePlaceholderString = (str) => {
 };
 
 export const getAllDescendants = (node) => {
-    if (!node.children) return [node.value];
+    if (!node.children || node?.children?.length === 0) return [node.value];
```
MaxwellKendall, line 60: Adding extra case here to ensure that we're also returning `[node.value]` if the children prop is defined, but actually just an empty array.

src/js/helpers/checkboxTreeHelper.js
```diff
@@ -355,19 +355,44 @@ export const expandNodeAndAllDescendantParents = (nodes, propForNode = 'value',
...
+ * (a) the nodes from search results
+ * (b) the pre-existing nodes not apart of search results that are hidden via the className obj property w/ the value "hide"
+ * (c) placeholder nodes to represent the presence of partial children
+*/
+export const mergeChildren = (parentFromSearch, existingParent) => {
     if (existingParent.children && parentFromSearch.children) {
+        // Hide all existing children by default.
         const existingChildArray = existingParent
             .children
-            .filter((node) => {
```
MaxwellKendall, line 364: This was handling a case that actually will never exist.

src/js/containers/search/filters/naics/NAICSContainer.jsx
```diff
@@ -197,19 +197,24 @@ export class NAICSContainer extends React.Component {
...
-            this.state.stagedNaicsFilters,
-            this.props.nodes
-        );
-
-        this.props.setUncheckedNaics(newUnchecked);
-        this.props.stageNaics(newChecked, newUnchecked, stagedNaicsFilters);
-        this.props.setCheckedNaics(newChecked);
-
-        this.setState({ stagedNaicsFilters });
+        if (uncheckedNode.checked) {
```
jonhill13, line 200: I guess this is for if the unchecked is a parent?

MaxwellKendall: Kind of weird behavior; the checkboxTree invokes `onUncheck` when you check a half-checked node. But it does convey the checked state is still `checked: true`; whereas if it were a full check, then an uncheck, `checked: false` would be the state of this object.

src/js/helpers/checkboxTreeHelper.js
jonhill13, line 60: do we need this `node?.children?.length` couldn't it be `node?.children.length` since children is always an array or has that changed?

MaxwellKendall: children is always an array. that hasn't changed. but I dont think the chain operator makes children an array if it's undefined, so `undefined/value-of-chain-operator-when-prop-does-not-exist.length` probably won't work w/o another chain operator.

MaxwellKendall: node.children though could be absent from the object altogether.

jonhill13: yeah, I think I have been reading it wrong from when I started using it. Thanks

PR #1763, opened 2020-06-24
Bug & Updates Contract Grant Activity Chart

jonhill13:
**High level description:**

> • fixes a bug where the action type in the tooltip was displaying the type description.

> • Updates

Given a start date, no end date and one transaction:

• only chart when start date is not equal to transaction date

https://federal-spending-transparency.atlassian.net/browse/DEV-5479

https://federal-spending-transparency.atlassian.net/browse/DEV-4886

- [x] Code review complete

### DIFF
--- a/src/js/components/sharedComponents/DatePicker.jsx
+++ b/src/js/components/sharedComponents/DatePicker.jsx
@@ -234,8 +234,7 @@ export default class DatePicker extends React.Component {
                     <a
                         href="#null"
                         onClick={this.toggleDatePicker}
-                        className="usa-da-icon picker-icon date"
-                        aria-haspopup="true">
+                        className="usa-da-icon picker-icon date">
                         <Icons.Calendar alt="Date picker" />
                     </a>
                 </div>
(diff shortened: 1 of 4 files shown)


src/js/components/sharedComponents/DatePicker.jsx
```diff
@@ -234,8 +234,7 @@ export default class DatePicker extends React.Component {
                     <a
                         href="#null"
                         onClick={this.toggleDatePicker}
-                        className="usa-da-icon picker-icon date"
-                        aria-haspopup="true">
```
ebdabbs, line 238: Was this throwing an error in axe? It makes sense to me that we would want to indicate to assistive tech that this link opens a popup.

tests/helpers/contractGrantActivityHelper-test.js
```diff
@@ -73,9 +73,9 @@ describe('Contract Grant Activity Helper', () => {
                 expect(areTransactionDatesOrAwardAmountsInvalid(badStartDate, 'grant', newTransaction)).toBe(true);
             });
             // 5/b
-            it('should return true when there is no end date and start date exists and only one transaction and transaction date is < than start date', () => {
+            it('should return true when there is no end date and start date exists and only one transaction and transaction date is === than start date', () => {
```
ebdabbs, line 76:
```suggestion
            it('should return true when there is no end date and start date exists and only one transaction and transaction date is equal to start date', () => {
```

src/js/components/sharedComponents/DatePicker.jsx
jonhill13: yeah, it randomly is in all of my branches, linter error, it will not build with that there.

PR #1835, opened 2020-07-17
DEV-5602 & DEV-5629 COVID recipient search bar and links update

ebdabbs:
**High level description:**
Adds a search bar to the COVID Profile **Spending by Recipient** section > **Recipients** tab table

**Technical details:**

- Adds `query` param to API request when the search string is more than 3 characters
- Creates a `SearchBar` component that could be reused in other sections of the page
- Styles substring of the recipient name that matches the search input
- When only one recipient profile page is available, link the recipient name instead of using `as Child`/ `as Recipient`
- Fixes a bug causing the Loans and Direct Payment tab counts to be swapped

**JIRA Ticket:**
[DEV-5602](https://federal-spending-transparency.atlassian.net/browse/DEV-5602) (search bar), [DEV-5629](https://federal-spending-transparency.atlassian.net/browse/DEV-5629) (recipient links)

**Mockup:**
https://bahdigital.invisionapp.com/share/TPIAEAE7HVG#/296042274_COVID-_Recipients_-_Recipient_Location_-_Recipient

Author:
- [x] Linked to this PR in JIRA tickets

Reviewer(s):
- [x] Design review complete @AGuyNamedMarco & @marcyheld 
- [ ] Code review complete

### DIFF
--- a/src/js/components/covid19/SearchBar.jsx
+++ b/src/js/components/covid19/SearchBar.jsx
@@ -8,12 +8,12 @@ import PropTypes from 'prop-types';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 const propTypes = {
-    setQuery: PropTypes.func.isRequired,
-    query: PropTypes.string
+    setQuery: PropTypes.func.isRequired
 };
 
-const SearchBar = ({ setQuery, query }) => {
-    const [searchString, setSearchString] = useState(query);
+const SearchBar = ({ setQuery }) => {
+    const [currentSearchTerm, setCurrentSearchTerm] = useState('');
+    const [searchString, setSearchString] = useState('');
 
     const onChange = (e) => {
         setSearchString(e.target.value);
@@ -21,6 +21,7 @@ const SearchBar = ({ setQuery, query }) => {
 
     const onSubmit = () => {
         setQuery(searchString);
+        setCurrentSearchTerm(searchString);
     };
 
     const resetSearch = () => {
@@ -29,7 +30,7 @@ const SearchBar = ({ setQuery, query }) => {
     };
 
     const handleClick = () => {
-        if (searchString && query === searchString) {
+        if (searchString && currentSearchTerm === searchString) {
             resetSearch();
         }
         else {
@@ -49,7 +50,7 @@ const SearchBar = ({ setQuery, query }) => {
                 aria-label="Search"
                 onClick={handleClick}
                 className="search-bar__button">
-                <FontAwesomeIcon icon={(searchString && query === searchString) ? 'times' : 'search'} />
+                <FontAwesomeIcon icon={(searchString && currentSearchTerm === searchString) ? 'times' : 'search'} />
             </button>
         </form>
     );
--- a/src/js/components/covid19/SearchBar.jsx
+++ b/src/js/components/covid19/SearchBar.jsx
@@ -1,0 +1,63 @@
+/**
+ * SearchBar.jsx
+ * Created by Lizzie Salita 7/16/20
+ */
+
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+const propTypes = {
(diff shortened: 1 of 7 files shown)


src/js/components/covid19/SearchBar.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+const SearchBar = ({ setQuery, query }) => {
+    const [searchString, setSearchString] = useState(query);
+
+    const onChange = (e) => {
+        setSearchString(e.target.value);
+    };
+
+    const onSubmit = (e) => {
+        e.preventDefault();
```
jonhill13, line 23: Any reason we are passing the event here?

src/js/components/covid19/SearchBar.jsx
```diff
@@ -0,0 +1,61 @@
...
+        setQuery(searchString);
+    };
+
+    const resetSearch = (e) => {
```
jonhill13, line 27: Any reason we are passing the event here?

src/js/components/covid19/SearchBar.jsx
```diff
@@ -0,0 +1,61 @@
...
+        e.preventDefault();
+        setSearchString('');
+        setQuery('');
+    };
+
+    const handleClick = (e) => {
+        if (searchString && query === searchString) {
+            resetSearch(e);
```
jonhill13, line 35: Any reason we are passing the event here?

src/js/components/covid19/SearchBar.jsx
```diff
@@ -0,0 +1,61 @@
...
+        }
+        else {
+            onSubmit(e);
```
jonhill13, line 38: Any reason we are passing the event here?

src/js/components/covid19/SearchBar.jsx
```diff
@@ -0,0 +1,61 @@
...
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+const propTypes = {
+    setQuery: PropTypes.func.isRequired,
+    query: PropTypes.string
+};
+
+const SearchBar = ({ setQuery, query }) => {
```
jonhill13, line 15: Any reason we are passing query here?

ebdabbs: `query` (what the table is currently filtered on) is compared to `searchString` (value of the input) to determine whether `handleClick` will reset or perform a new search (and to display the corresponding icon)

jonhill13: Yes, I guess I am confused? The search bar should be taking input and searching. I do not understand why search bar needs a parent to pass it the search term? There must be a case you are trying to handle here, I do not think I see it ( possibly when a user types the same input again? ).

ebdabbs: Right, if their input matches the existing search it prevents submitting the same search term again and instead gives them the option to reset.

jonhill13: Yeah, it just seems like this component should be self sufficient. The search bar should be able to handle that logic itself.  We could remove the query param and access the input element onSubmit to get the value or move that query parameter to the search bar like

```const SearchBar = ({ setQuery }) => {
    const [searchString, setSearchString] = useState('');
    const [currentSearchTerm, setCurrentSearchTerm] = useState('');

    const onChange = (e) => {
        setSearchString(e.target.value);
    };

    const onSubmit = (e) => {
        e.preventDefault();
        setCurrentSearchTerm(searchString);
        setQuery(searchString);
    };

    const resetSearch = (e) => {
        e.preventDefault();
        setSearchString('');
        setQuery('');
    };

    const handleClick = (e) => {
        if (searchString && currentSearchTerm === searchString) {
            resetSearch(e);
        }
        else {
            onSubmit(e);
        }
    };```

Thus, in the future when other devs implement this I need to do less work by just passing setQuery

`<SearchBar setQuery={search} />`

and not worry about handling the query param.

PR #1843, opened 2020-07-20
[DEV-5197] Preventing onSubmit triggered by onClick of disabled Quarter

MaxwellKendall:
**High level description:**
There is a bug where clicking a disabled quarter on the account download page triggers the download.

**Technical details:**
This was happening because in the component from the library, we are not preventing an onClick event on the Quarter Picker from bubbling up to the form and creating false positive for the onSubmit event.

Rather than fix the component library now, I've changed our implementation such that we aren't looking for an onSubmit event on the form generally, but rather we're attaching an onClick event to the submit button itself. This avoids the problem. [I created a ticket for the component library so we can fix the root cause later](https://federal-spending-transparency.atlassian.net/browse/DEV-DEV-5687).

**JIRA Ticket:**
[DEV-5197](https://federal-spending-transparency.atlassian.net/browse/DEV-5197)

Author:
`N/A` Verified cross-browser compatibility: Chrome, Safari, Firefox, Internet Explorer, Edge
`N/A` Added Unit Tests for methods in Container Components, reducers, helper functions, and models `if applicable`
`N/A` All `componentWillReceiveProps` and `componentWillMount` in relevant child/parent components/containers replaced with [future compatible life-cycle methods](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) per [JIRA item 3277](https://federal-spending-transparency.atlassian.net/browse/DEV-3277)

Reviewer(s):
- [x] Code review complete

### DIFF
--- a/src/js/components/bulkDownload/accounts/AccountDataContent.jsx
+++ b/src/js/components/bulkDownload/accounts/AccountDataContent.jsx
@@ -55,9 +55,7 @@ export default class AccountDataContent extends React.Component {
         this.props.clearAccountFilters();
     }
 
-    handleSubmit(e) {
-        e.preventDefault();
-
+    handleSubmit() {
         this.props.clickedDownload();
     }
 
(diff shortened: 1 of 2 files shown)


MaxwellKendall:
Had to add 51b2a67 because clicking the button would update the url... I confirmed this change:

1. Has no effect on styling
2. Has no effect on accessibility.

If we prefer we can fix this by updating the component library, but I think it's unnecessary at the moment given priority etc.

src/js/components/bulkDownload/accounts/AccountDataContent.jsx
```diff
@@ -55,9 +55,7 @@ export default class AccountDataContent extends React.Component {
         this.props.clearAccountFilters();
     }
 
-    handleSubmit(e) {
-        e.preventDefault();
-
+    handleSubmit() {
```
ebdabbs, line 58: Looks like we could get rid of this function and just pass `this.props.clickedDownload` as to `SubmitButton`.

PR #1941, opened 2020-07-31
QAT -- ie style changes

parikh-adit:
-fix internet explorer bug


src/_scss/layouts/default/header/nav/_dropdownMenu.scss
```diff
@@ -21,7 +21,7 @@
         font-weight: $font-semibold;
         font-size: rem(15);
         line-height: rem(19);
-        display: contents;
+        display: flex;
```
MaxwellKendall, line 24:
```suggestion
        @include display(flex);
```

PR #2057, opened 2020-09-03
[DEV-5401] spending explorer periods

emily-brents:
**High level description:**

Warmfix adding support for defaulting to the latest period in the spending explorer rather than the latest quarter. Also adjusts some of the css.

**JIRA Ticket:**
[DEV-5401](https://federal-spending-transparency.atlassian.net/browse/DEV-5401)

**Mockup:**
https://bahdigital.invisionapp.com/share/C8IADXT2BD4#/screens/296036477

Author:
- [X] Linked to this PR in JIRA ticket
- [X] Verified cross-browser compatibility: Chrome, Safari, Firefox, Internet Explorer, Edge
- [X] Verified mobile/tablet/desktop/monitor responsiveness
- [X] Verified that this PR does not create any *new* accessibility issues (via Axe Chrome extension)
- [X] Added Unit Tests for methods in Container Components, reducers, helper functions, and models `if applicable`
- [X] All `componentWillReceiveProps` and `componentWillMount` in relevant child/parent components/containers replaced with [future compatible life-cycle methods](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) per [JIRA item 3277](https://federal-spending-transparency.atlassian.net/browse/DEV-3277)

Reviewer(s):
- [ ] Design review complete
- [ ] Code review complete


MaxwellKendall (changes requested):
I'm gonna be out tomorrow but I think the main thing that needs to be looked at is this `mostRecentPeriod` fn. Not sure why we can't use [`getInitialPeriodByFy`](https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/redux/reducers/bulkDownload/bulkDownloadReducer.js#L14)?

Also the comments or the code need to be updated. I don't understand the logic behind how we're determining the active period. If for some reason we can't use the existing `fn` that selects the default period by FY in the bulkDownloadReducer, can we just update the comments so the logic is more readable?

Thanks for diving into the front end!

src/js/containers/explorer/detail/helpers/explorerQuarters.js
```diff
@@ -150,3 +150,48 @@ export const defaultQuarters = () => {
     const current = mostRecentQuarter();
     return availableQuartersInFY(current.year);
 };
+
+
+export const mostRecentPeriod = () => {
+    // go back 30 days prior to today
+    const todayAdjusted = moment().add(3, 'months');
```
MaxwellKendall, line 157: Update the comment? Not sure what the logic is we're using here.

MaxwellKendall: I'm guesing this should be 

```suggestion
    const todayAdjusted = moment().subtract(30, 'days');
```

src/js/redux/reducers/explorer/explorerReducer.js
```diff
@@ -14,12 +14,12 @@ export const ActiveScreen = new Record({
 });
 
 const initialQuarters = defaultQuarters();
-
+const initalPeriods = defaultPeriod();
```
MaxwellKendall, line 17: We should be able to just use [this fn in the bulkDownloadReducer](https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/redux/reducers/bulkDownload/bulkDownloadReducer.js#L14) for this.

emily-brents: We can't, the `getInitialPeriodByFy` function gets the latest period in the lastest closed quarter, which is not the latest closed period if that makes sense. Essentially the period will close before the quarter will, but that function is checking for closed quarters, which is what was causing the problem

src/js/containers/explorer/detail/helpers/explorerQuarters.js
emily-brents: yeah that's what i get for copy pasting, updated the comment to be more clear

emily-brents:
accidentally merged in dev

PR #2067, opened 2020-09-08
Fix dev 5819 covid19map

parikh-adit:
**High level description:**

PR created to fix the issue where counties or states provide minimal data and the map was not colored as expected. 

**Technical details:**

The map now uses a linear scale when we know there are fewer than 10 unique values. The linear scale works much better than a quantile scale in cases where data is minimal or not varied. 

**JIRA Ticket:**
[DEV-5819](https://federal-spending-transparency.atlassian.net/browse/DEV-5819)

Author:
- [x] Scheduled demo including Design/Testing/Front-end OR Provided instructions for testing in JIRA and PR `if applicable`
- [x] All `componentWillReceiveProps` and `componentWillMount` in relevant child/parent components/containers replaced with [future compatible life-cycle methods](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) per [JIRA item 3277](https://federal-spending-transparency.atlassian.net/browse/DEV-3277)
- [x] [API contract](https://github.com/fedspendingtransparency/usaspending-api/tree/dev/usaspending_api/api_contracts) updated `if applicable`
- [x] [Component Library Integration Status Report](https://github.com/fedspendingtransparency/data-act-documentation/blob/data-transparency-ui/frontend_apps/component-library-integration-status.md) updated `if applicable`

Reviewer(s):
- [x] Code review complete


parikh-adit:
https://federal-spending-transparency.atlassian.net/browse/DEV-6070

src/js/components/covid19/recipient/map/MapWrapper.jsx
```diff
@@ -376,20 +385,42 @@ export default class MapWrapper extends React.Component {
...
-        // calculate the range of data
-        const scale = calculateCovidMapRange(this.props.data.values, this.props.activeFilters.territory);
-        const colors = this.getColors();
+
+        // the d3 function takes a range array for the key/index for each quantile or segment in the case of a linear scale
+        let rangeArray = [];
+        let colors = [];
+        if (this.props.activeFilters.territory === 'state') {
+            colors = this.getColors(numStateQuantiles);
+            rangeArray = [...Array(numStateQuantiles).keys()];
```
jonhill13, line 396: do we need range array to be an array of indices? or do we just need an array with the same length as the colors array? if the latter, then we can just use the colors.

src/js/components/covid19/recipient/map/MapWrapper.jsx
```diff
@@ -376,20 +385,42 @@ export default class MapWrapper extends React.Component {
...
         const source = mapboxSources[this.props.activeFilters.territory];
-        // calculate the range of data
-        const scale = calculateCovidMapRange(this.props.data.values, this.props.activeFilters.territory);
-        const colors = this.getColors();
+
+        // the d3 function takes a range array for the key/index for each quantile or segment in the case of a linear scale
+        let rangeArray = [];
+        let colors = [];
+        if (this.props.activeFilters.territory === 'state') {
+            colors = this.getColors(numStateQuantiles);
```
jonhill13, line 395: Once colors is moved to state you can reference it here.

src/js/components/covid19/recipient/map/MapWrapper.jsx
```diff
@@ -209,7 +210,16 @@ export default class MapWrapper extends React.Component {
 
         // generate the highlight layers that will be shaded in when populated with data filters
         // set up temporary empty filters that will show nothing
-        const colors = this.getColors();
+        let colors = [];
+        if (this.props.data.values.length !== 0) {
+            if (this.props.activeFilters.territory === 'state') {
+                colors = this.getColors(numStateQuantiles);
```
jonhill13, line 216: lets add colors to state since we are doing this multiple times.

jonhill13 (changes requested):

jonhill13:
@parikh-adit did we get a design review on this?

src/js/components/covid19/recipient/map/MapWrapper.jsx
parikh-adit: it needs to be an array of indices because the D3 scale functions map a domain to a range, so the rangeArray cannot be empty. The numbers in the range array [0, 1, 2....] correspond to to the layers that are added to the map. Similar to what we were doing when there were only 6 layers we used to color the map.

src/js/components/covid19/recipient/map/MapWrapper.jsx
parikh-adit: I can make two arrays in the state for the class one for the county colors and state colors but those conditional statements will remain to get the appropriate rangeArray based on this.props.activeFilters.territory. Unfortunately, I could not figure out how to render the map without hard coding these values. 

The issue is because in the loadSource() function there is we set up sets of empty filters based on this.props.activeFilters.territory that can later be used to color the map in displayData(). In cases when we had limited reported data because of the selected filters, this was causing issues. 

For example, say we land on the page and get a map with all 50 states colored. Then we select filters such that we only have data for 15 states the behaviour was not as expected because we had previously generated a set of 50 filters and only the first 15 are used. I tried moving out the the code in the loadSource() function that generates the sets of empty filters so in cases like the one I described the map would have a set of 15 colors to use. Im sure this is the most effective method to refactor the code to account for smaller datasets with missing values for some states. However, the when I tried doing it I was creating errors and then there was a lag in rendering the map and then the page would re-render but abnormally, with like a quarter of the map having no color.

jonhill13 (approved):

PR #2094, opened 2020-09-15
[DEV-5551][DEV-5549] Fix wording and some award counts

sethstoudenmier:
**High level description:**

Update unlinked text to match AC and use Award Count when determining missing awards.

**Technical details:**

Refactored some to minimize calls to `/api/v2/disaster/award/amount/` since the endpoint is causing a lot of trouble. Instead of multiple calls for the same endpoint there is one call to handle the "All Awards" tab across different visualizations and that one call is stored in the Redux store.

**JIRA Ticket:**
[DEV-5551](https://federal-spending-transparency.atlassian.net/browse/DEV-5551)
[DEV-5549](https://federal-spending-transparency.atlassian.net/browse/DEV-5549)

**Mockup:**
N/A

Author:
- [x] All `componentWillReceiveProps` and `componentWillMount` in relevant child/parent components/containers replaced with [future compatible life-cycle methods](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) per [JIRA item 3277](https://federal-spending-transparency.atlassian.net/browse/DEV-3277)

Reviewer(s):
- [ ] Code review complete

### DIFF
--- a/src/js/components/covid19/assistanceListing/SpendingByCFDA.jsx
+++ b/src/js/components/covid19/assistanceListing/SpendingByCFDA.jsx
@@ -61,6 +61,13 @@ const SpendingByCFDA = () => {
         Analytics.event({ category: 'COVID-19 - Award Spending by CFDA', action: `${activeTab} - click` });
     };
 
+    // Keep track of previous tab to prevent duplicate requests in summary
+    const prevTabRef = useRef();
+    useEffect(() => {
+        prevTabRef.current = activeTab;
+    });
+    const prevTab = prevTabRef.current;
+
     useEffect(() => {
         if (defCodes && defCodes.length > 0) {
             // Make an API request for the count of CFDA for each award type
@@ -133,6 +140,7 @@ const SpendingByCFDA = () => {
                 // pass CFDA count to the summary section so we don't have to make the same API request again
                 resultsCount={tabCounts[activeTab]}
                 activeTab={activeTab}
+                prevTab={prevTab}
                 areCountsLoading={inFlight}
                 overviewData={overviewData}
                 assistanceOnly />
--- a/src/js/containers/covid19/Covid19Container.jsx
+++ b/src/js/containers/covid19/Covid19Container.jsx
@@ -37,8 +37,13 @@ import {
     stickyHeaderHeight,
     dataDisclaimerHeight
 } from 'dataMapping/covid19/covid19';
-import { fetchDEFCodes, fetchOverview, fetchAllSubmissionDates } from 'helpers/disasterHelper';
-import { setDEFCodes, setOverview, setLatestSubmissionDate } from 'redux/actions/covid19/covid19Actions';
+import {
+    fetchDEFCodes,
+    fetchOverview,
+    fetchAllSubmissionDates,
+    fetchAwardAmounts
+} from 'helpers/disasterHelper';
+import { setDEFCodes, setOverview, setLatestSubmissionDate, setTotals } from 'redux/actions/covid19/covid19Actions';
 import { showModal } from 'redux/actions/modal/modalActions';
 import DataSourcesAndMethodology from 'components/covid19/DataSourcesAndMethodology';
 import OtherResources from 'components/covid19/OtherResources';
@@ -58,6 +63,7 @@ const Covid19Container = () => {
     const defCodesRequest = useRef(null);
     const overviewRequest = useRef(null);
     const lastSectionRef = useRef(null);
+    const awardAmountRequest = useRef(null);
     const dataDisclaimerBannerRef = useRef(null);
     const allSubmissionDatesRequest = useRef(null);
     const dispatch = useDispatch();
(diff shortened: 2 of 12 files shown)


src/js/components/covid19/assistanceListing/SpendingByCFDA.jsx
```diff
@@ -61,6 +61,13 @@ const SpendingByCFDA = () => {
         Analytics.event({ category: 'COVID-19 - Award Spending by CFDA', action: `${activeTab} - click` });
     };
 
+    // Keep track of previous tab to prevent duplicate requests in summary
+    const prevTabRef = useRef();
+    useEffect(() => {
+        prevTabRef.current = activeTab;
```
MaxwellKendall, line 67: This duplicates our state. A `ref` [should be used sparingly for `mutuable` data](https://reactjs.org/docs/hooks-reference.html#useref). Typically this means a DOM Node. This is how we use it. In this case, we're putting immutable data that duplicates a property already in state inside a ref. We should just use the state property here.

src/js/components/covid19/assistanceListing/SpendingByCFDA.jsx
```diff
@@ -61,6 +61,13 @@ const SpendingByCFDA = () => {
...
+    });
+    const prevTab = prevTabRef.current;
```
MaxwellKendall, line 69: In this case, we're setting the `prevTab` to equal the value of a ref, which itself is the value of a property in state called `activeTab`. So `prevTab` equals `activeTab`? This is self-refuting and confusing. Not sure what exactly we're trying to do here, but I think we're `trying` to capture stale state? This shouldn't be necessary.

src/js/containers/covid19/SummaryInsightsContainer.jsx
```diff
@@ -43,53 +45,71 @@ const SummaryInsightsContainer = ({
     const [numberOfAwards, setNumberOfAwards] = useState(null);
     const initialInFlightState = overviewData.map((d) => d.type);
     const [inFlightList, , removeFromInFlight, resetInFlight] = useInFlightList(initialInFlightState);
-    const defCodes = useSelector((state) => state.covid19.defCodes);
+    const { defCodes, allAwardTypeTotals } = useSelector((state) => state.covid19, isEqual);
 
     useEffect(() => {
+        if (assistanceOnly && (activeTab === 'all' && prevTab === 'all')) {
```
MaxwellKendall, line 51: This reminds me of using component lifecycle methods. W/ useEffect, we have to use a different mental model. I'm not confident this prevTab value is going to be reliable.

Also, are we making duplicative requests to this endpoint in production?  If so, that could be a bug you can scope outside this delivery.

src/js/helpers/covid19Helper.js
```diff
@@ -164,10 +164,13 @@ export const getTotalSpendingAbbreviated = (totalSpending) => {
...
-export const areCountsDefined = (counts) => Object.keys(counts).reduce((acc, tab) => {
-    if (acc === null) return acc;
-    return counts[tab];
-}, true);
+export const areCountsDefined = (counts) => {
+    const countTabs = Object.keys(counts);
+    if (countTabs.length === 0) {
+        return false;
+    }
+    return countTabs.reduce((acc, tab) => acc && counts[tab] !== null, true);
```
MaxwellKendall, line 172: Can we add a test case to cover the condition where this was returning `falsy`? This would just mean adding an `it` block to the existing tests where one of the tab counts is zero; preferably the last property in the tabCounts object like

```javascript
const mockTabCounts = { test1: 5, test2: 1, test3: 0 }
```

src/js/containers/covid19/Covid19Container.jsx
```diff
@@ -121,14 +127,39 @@ const Covid19Container = () => {
...
         };
+        const getAllAwardTypesAmount = async () => {
+            const params = {
+                filter: {
+                    def_codes: defCodes.map((code) => code.code)
+                }
+            };
+            awardAmountRequest.current = fetchAwardAmounts(params);
+            awardAmountRequest.current.promise
+                .then((res) => {
```
MaxwellKendall, line 138: to pass the eslint camelcase r ule:

```suggestion
                .then(({
                  data: {
                    obligation,
                    outlay,
                    award_count: awardCount,
                    face_value_of_loan: faceValueOfLoan
                  }
                }) => {
```

src/js/containers/covid19/SummaryInsightsContainer.jsx
```diff
@@ -43,53 +45,71 @@ const SummaryInsightsContainer = ({
...
+                    params.filter.award_type = 'assistance';
+                }
+                awardAmountRequest.current = fetchAwardAmounts(params);
+                awardAmountRequest.current.promise
+                    .then((res) => {
+                        setAwardObligations(res.data.obligation);
+                        setAwardOutlays(res.data.outlay);
+                        setNumberOfAwards(res.data.award_count);
+
+                        /* eslint-disable camelcase */
```
MaxwellKendall, line 93: ditto per above we can use object de-structuring to pass this test and we probably should to make it more readable as to what these properties on the api actually are.

MaxwellKendall (changes requested):

src/js/components/covid19/assistanceListing/SpendingByCFDA.jsx
sethstoudenmier: This particular use case I think makes sense to not use state sense I am trying to track the previous value of the state: https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state

To use state would mean adding an additional update in the code for each state update when you can let this ref handle tracking that.

src/js/components/covid19/assistanceListing/SpendingByCFDA.jsx
sethstoudenmier: https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state

src/js/containers/covid19/SummaryInsightsContainer.jsx
sethstoudenmier: As mentioned in our quick call, this particular duplicate call was noticed as a result of "all" for CFDA being considered something different than the other tables. Want to make sure we aren't sending a second request for CFDA when an update to `allAwardTypeTotals` triggers this hook.

src/js/helpers/covid19Helper.js
sethstoudenmier: Will do

src/js/components/covid19/assistanceListing/SpendingByCFDA.jsx
MaxwellKendall: Oh my! I had no idea this was a convention recommended in the documentation. Nicely done sir.

sethstoudenmier: I will say if this becomes a more common thing then we should create a custom hook, but right now this is just a one off use case.

MaxwellKendall (approved):

PR #2189, opened 2020-10-22
Fix Merge Conflicts Master to QAT

jonhill13:
- [ ] Code review complete

### DIFF
--- a/src/js/containers/covid19/awardSpendingAgency/AwardSpendingAgencyTableContainer.jsx
+++ b/src/js/containers/covid19/awardSpendingAgency/AwardSpendingAgencyTableContainer.jsx
@@ -269,7 +269,7 @@ const AwardSpendingAgencyTableContainer = (props) => {
     });
 
     useEffect(() => {
-        if (!Object.keys(spendingByAgencyTotals).length && resultsTotal) {
+        if (Object.keys(spendingByAgencyTotals).length && resultsTotal) {
             addUnlinkedData(results, resultsTotal, spendingByAgencyTotals);
         }
     }, [spendingByAgencyTotals, resultsTotal]);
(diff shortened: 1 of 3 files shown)


src/js/containers/covid19/awardSpendingAgency/AwardSpendingAgencyTableContainer.jsx
```diff
@@ -269,7 +269,7 @@ const AwardSpendingAgencyTableContainer = (props) => {
     });
 
     useEffect(() => {
-        if (!Object.keys(spendingByAgencyTotals).length && resultsTotal) {
+        if (Object.keys(spendingByAgencyTotals).length && resultsTotal) {
```
MaxwellKendall, line 272: if this is putting commits from master into qat, there should be about 3 other fiels where this same change is also put into qat.

MaxwellKendall (changes requested):

src/js/containers/covid19/awardSpendingAgency/AwardSpendingAgencyTableContainer.jsx
MaxwellKendall: What happened? These files are the ones I changed that should show up here. (From your closed PR merging master into qat)

image

jonhill13: Yeah, I did not want to include all the changes from the down merge PR.  I only want to fix the merge conflicts. Which I thought would only be in the package.json. But, there were conflicts in this container because of the refactoring I did of the render function of all of these award sections. So I had to fix that, and this change came with it from master.

The down merge should bring all those other changes and you should be able to see them in down merge PR.

MaxwellKendall (approved):

PR #2227, opened 2020-11-18
Dev 6204 add download to advanced time tab

brett-varney:
**High level description:**

Adds feature to download current chart data to CSV.

**Technical details:**

Adds text button to download to UI & info icon/text per mock
Adds code to visualization to convert current chart data to CSV and download when button clicked.

**JIRA Ticket:**
https://federal-spending-transparency.atlassian.net/browse/DEV-5595

**Mockup:**
https://bahdigital.invisionapp.com/share/ACIAFALFPGE#/screens

Author:
- [x] Verified cross-browser compatibility: Chrome, Safari, Firefox, Internet Explorer, Edge (author uses Windows, so cannot check Safari, but others all works properly in Windows)
- [x] Verified mobile/tablet/desktop/monitor responsiveness (page/chart not currently responsive, but feature remains in-place inside containers)

Reviewer(s):
- [ ] Code review complete


src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
```diff
@@ -16,6 +20,12 @@ const propTypes = {
     visualizationPeriod: PropTypes.string
 };
 
+const downloadTooltip = (
+    <div className="tooltip__text">
```
MaxwellKendall, line 24: We should reuse styles from other tooltips for consistency. Likely the designers are gonna flag this. We have a TooltipComponent exported from the component library you can try and use and then append the right class  names or jsut us a div with the right class name and mark up structure.

See `customTooltipContent.scss` for relelvant classnames.

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
```diff
@@ -49,6 +59,64 @@ export default class TimeVisualizationSection extends React.Component {
         }
     }
 
+    downloadLabel = () => {
```
MaxwellKendall, line 62: This can just be a ternary  inside the render.

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
```diff
@@ -49,6 +59,64 @@ export default class TimeVisualizationSection extends React.Component {
...
+        if (data.visualizationPeriod === 'fiscal_year') {
+            ret[0] = ['fiscal_year', 'total_obligations'];
+        }
+        else if (data.visualizationPeriod === 'quarter') {
+            ret[0] = ['fiscal_year', 'fiscal_quarter', 'total_obligations'];
+        }
+        else {
+            ret[0] = ['fiscal_year', 'month', 'total_obligations'];
+        }
+        for (let i = 0; i < data.rawLabels.length; i++) {
```
MaxwellKendall, line 85: this should be a `.map`

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
```diff
@@ -49,6 +59,64 @@ export default class TimeVisualizationSection extends React.Component {
...
+        else {
+            periodLabel = capitalize(this.props.data.visualizationPeriod);
+        }
+        return `Download data by ${periodLabel}`;
+    };
+
+    downloadData = () => {
+        const data = this.props.data;
+        const ret = [];
+        if (data.visualizationPeriod === 'fiscal_year') {
```
MaxwellKendall, line 76: we should use an object keyed on `data.visualsationPeriod` w/ the arrays as the value and turn these ifs elses into one line:

```javascript
const visualizationPeriodByPeriod {
   fiscal_year: ['some', 'strings']
};

data.visualizationPeriod = visualizationPeriodByPeriod[data.visualizationPeriod]
```

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
```diff
@@ -49,6 +59,64 @@ export default class TimeVisualizationSection extends React.Component {
...
+        if (this.props.data.visualizationPeriod === 'fiscal_year') {
+            periodLabel = 'Year';
+        }
+        else {
+            periodLabel = capitalize(this.props.data.visualizationPeriod);
+        }
+        return `Download data by ${periodLabel}`;
+    };
+
+    downloadData = () => {
```
MaxwellKendall, line 73: can we rename this to describe what's being returned?

like `get...?`

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
```diff
@@ -49,6 +59,64 @@ export default class TimeVisualizationSection extends React.Component {
...
+                    // correct FY
+                    ret[i + 1][0] = parseInt(ret[i + 1][0], 10) + 1;
+                }
+            }
+        }
+        return ret;
+    };
+
+    downloadCsv = () => {
+        let contents = this.downloadData();
```
MaxwellKendall, line 105: we shouldn't use `let` here. Mutable data is troublesome.

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
```diff
@@ -49,6 +59,64 @@ export default class TimeVisualizationSection extends React.Component {
...
+        contents = contents.map((row) => row.join(',')).join('\n');
+        // eslint-disable-next-line no-undef
+        const file = new Blob([contents], { type: 'text/csv;charset=utf-8;' });
+        if (isIe()) {
+            window.navigator.msSaveOrOpenBlob(file, 'spending-over-time.csv');
+        }
+        else {
+            const a = document.createElement('a');
```
MaxwellKendall, line 113: We should just put this in the render, hide it, set the href to a value in state, and set state here. We don't want to avoid using the react conventions. We have all these listeners attached to the document already so we do our best not to interface w/ react.

We should also likely provide a consistent UX for download and present the user with the downloadBottomBar, providing them w/ a link that onClick will download.

We can pair on this and should take like 10min.

src/js/helpers/general.js
```diff
@@ -0,0 +1,2 @@
+// check if browser is IE
+export const isIe = () => !!document.documentMode;
```
MaxwellKendall, line 2: can we rename this to helper/browser.js?

src/js/helpers/monthHelper.js
```diff
@@ -39,3 +39,21 @@ export const convertPeriodToDate = (period, fy) => {
     const year = (parsedPeriod <= 3) ? parsedFy - 1 : parsedFy;
     return `${month} ${year}`;
 };
+
+export const fullMonthFromAbbr = (abbr) => {
```
MaxwellKendall, line 43: this should just be a dataMapping object. We might also already be doing this elsewhere.

```javascript
  const monthByMonthAbbreviation = {
      'Jan': 'January'
      // etc...
  }
  const month = monthByMonthAbbreviation[abbr];
```

MaxwellKendall (changes requested):
Let me know if you have any pushback. Glad to discuss my requests. Lets also do some pair on this for some changes that aren't as straight forward. Thanks for contributing!

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: What would you suggest? What's being returned is the data for the download. True, function names should be verbs, and this doesn't actually produce the download...

src/js/helpers/monthHelper.js
brett-varney: I searched the code, didn't see this specific conversion in this project. Maybe in an imported one...

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: I had that originally, but it looked long and awkward to read. But I'll put it back to follow project standards.

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: Yeah, this is unconventional, but since it needs to deal with two radically different approaches (thank you, IE), I don't see how to include both in a different way, but am open to your idea. I don't know what a downloadBottomBar is, but the name sounds misleading in this context.

src/js/helpers/general.js
brett-varney: done

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: true

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: Sorry, missed your last line here...

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: done

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: done

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: done

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: done

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
brett-varney: Since none of the following lines are simple string assignments, I am just not picturing how this would work. Let's chat.

src/js/components/search/visualizations/time/TimeVisualizationSection.jsx
MaxwellKendall: It's a part of our UX. This will have to go through design review and I anticipate they will want a consistent UX for download behavior on this tab. You'll see it and understand if you download a search on the first tab.

MaxwellKendall: We should be using react to render this a tag.  the href value should point at state and this should update state.

MaxwellKendall (changes requested):

jonhill13 (approved):

PR #2291, opened 2020-12-30
Hotfix DEV-6575 map error

ebdabbs:
**High level description:**

Fixes a bug where in rare edge cases, the Advanced Search map would throw an error resulting in a blank screen. E.g. https://www.usaspending.gov/search/13bd5bf59da6f9c844a8b20bc364895a. 

**Technical details:**

`MapWrapper` groups dollar values into 6 ranges represented by different colors on the map using the helper function `calculateRanges`. 

Old (flawed) approach: 

1. Determine the upper bound for our legend by rounding up the maximum dollar value in our data to an integer that is divisible by 6 * `unit` for the range of values (thousand, million, etc.)
2. Use d3 `scaleLinear` to convert each dollar value to a floating point number between 0 and 6
3. Use `Math.floor` to round down to the index of an array representing that bucket
4. Push the corresponding location to the array at the given index

This falls apart when the maximum dollar value in the data is a nice round number that is divisible by 6 * `unit` ($1,200,000 at the link above) and matches the upper bound used for the domain. In this case, `scaleLinear` will give us `6` as the index (which is correct), but there are only 6 buckets (the highest index of which is `5`).

Solution:
We were implementing what d3 `scaleQuantize` is designed to do; switch from `scaleLinear` to `scaleQuantize`. See https://github.com/d3/d3-scale/blob/master/README.md#quantize-scales

**JIRA Ticket:**
[DEV-6575](https://federal-spending-transparency.atlassian.net/browse/DEV-6575)

Author:

Reviewer(s):
- [ ] Code review complete

### DIFF
--- a/src/js/components/search/visualizations/geo/MapWrapper.jsx
+++ b/src/js/components/search/visualizations/geo/MapWrapper.jsx
@@ -420,8 +420,7 @@ export default class MapWrapper extends React.Component {
             let value = this.props.data.values[index];
             if (isNaN(value)) value = 0;
             // determine the group index
-            let group = Math.floor(scale.scale(value));
-            if (group.toString().startsWith('-')) group = 0;
+            const group = scale.scale(value);
             // add it to the filter list
             filterValues[group].push(location);
         });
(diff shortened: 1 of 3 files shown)


sethstoudenmier (approved):
LGTM. Everything seems to be working as expected. Color pattern appears to be the same between current deployed code vs the hotfix for filters that do not cause the page to go blank.

PR #2324, opened 2021-01-12
[DEV-6342] cleaning up broken styles on homepage tooltip

MaxwellKendall:
Changes in the component library created this bug: 

image

Fixing it:

image

- [ ] Code review complete


jonhill13:
@MaxwellKendall changed to hotfix

jonhill13:
looks like these tooltips are broke across the site? or just these with this specific class?
Screen Shot 2021-01-13 at 1 12 53 PM

MaxwellKendall:
I think its wherever we are removing padding/margin from the h2s:

src/_scss/components/tooltips/_customTooltipContent.scss

jonhill13 (approved):

PR #2344, opened 2021-01-21
[DEV-6536] Period 1 and Period 2 are combined, actually

MaxwellKendall:
**High level description:**
Mockup: https://bahdigital.invisionapp.com/d/main#/console/14222399/296054007/preview
image

- [x] Code review complete

### DIFF
--- a/src/js/containers/aboutTheData/AgencyTableMapping.jsx
+++ b/src/js/containers/aboutTheData/AgencyTableMapping.jsx
@@ -33,8 +33,8 @@ export const agenciesTableColumns = {
             displayName: 'FY 2020 Q1',
             columnSpan: "2",
             subColumnNames: [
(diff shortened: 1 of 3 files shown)


src/js/containers/aboutTheData/AgencyTableMapping.jsx
```diff
@@ -30,32 +30,31 @@ export const agenciesTableColumns = {
         { title: 'current_total_budget_authority_amount', displayName: 'Percent of Total Federal Budget' },
         {
             title: 'Q1',
-            displayName: 'FY 2020 Q1',
+            displayName: `FY 2020 Q1`,
```
jonhill13, line 33: We should use regular quotes if we do not to evaluate this string

jonhill13 (changes requested):

jonhill13 (approved):

PR #2368, opened 2021-02-01
DEV-6593 About the Data tooltips bug

ebdabbs:
**High level description:**

Prevents Unlinked Award column tooltips from getting cut off at the bottom of About the Data tables by making them wider.

**Technical details:**

The component library `wide` prop fills available space based on the width of the container. In this case, the width of the column itself caused the tooltip to actually get narrower when using that prop. This change overrides the width with USAS css using the `className` prop.

**JIRA Ticket:**
[DEV-6593](https://federal-spending-transparency.atlassian.net/browse/DEV-6593)

Author:
- [x] Scheduled demo including Design/Testing/Front-end OR Provided instructions for testing in JIRA and PR

Reviewer(s):
- [x] Code review complete

### DIFF
--- a/src/_scss/pages/aboutTheData/aboutTheData.scss
+++ b/src/_scss/pages/aboutTheData/aboutTheData.scss
@@ -56,4 +56,25 @@
             margin-right: rem(5);
         }
     }
+    .usda-table {
+        thead {
+            th {
+                .wide {
+                    .tooltip-spacer {
+                        width: 600px !important;
+                    }
+                    &.wide_left {
+                        .tooltip-spacer {
+                            left: -408px !important;
+                        }
+                    }
+                    &.wide-right {
+                        .tooltip-spacer {
+                            right: 408px !important;
+                        }
+                    }
+                }
+            }
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


src/_scss/pages/aboutTheData/aboutTheData.scss
```diff
@@ -56,4 +56,25 @@
...
+                        width: 600px !important;
+                    }
+                    &.wide_left {
+                        .tooltip-spacer {
+                            left: -408px !important;
+                        }
+                    }
+                    &.wide-right {
+                        .tooltip-spacer {
+                            right: 408px !important;
```
MaxwellKendall, line 73:
```suggestion
                    &.wide_left {
                        .tooltip-spacer {
                            height: rem(450);
                            overflow-y: scroll;
                        }
                    }
```

MaxwellKendall (changes requested):
We can adjust the height w/o using `important!`

src/_scss/pages/aboutTheData/aboutTheData.scss
ebdabbs: This line is adjusting the position `right` rather than height. I'm using important to override inline style coming from the component library.

ebdabbs: Were you suggesting scrolling w/in the tooltips instead? I can check with design (I'm personally not a fan of that UX)

MaxwellKendall (approved):

PR #2386, opened 2021-02-11
[DEV-6870] Hotfix DEFC "U" message

ebdabbs:
**High level description:**

Adds a message about DEFC "U" in two places.

**Technical details:**

* Includes unit tests for hiding this message on March 3, 2021.

**JIRA Ticket:**
[DEV-6870](https://federal-spending-transparency.atlassian.net/browse/DEV-6870)

Author:
- [x] Scheduled demo including Design/Testing/Front-end OR Provided instructions for testing in JIRA and PR 
  - Demo'd to @plooksm 
- [x] Added Unit Tests for helper functions, reducers, models and Container/Component Interactivity Expectations 

Reviewer(s):
- [x] Code review complete


tests/components/covid19/DataSourcesAndMethodologiesPage-test.jsx
```diff
@@ -0,0 +1,32 @@
...
+import { render, screen } from '@test-utils';
+
+const nativeDate = Date.now;
+
+afterAll(() => {
+    // restore the native date function
+    Date.now = nativeDate;
+});
+
+describe('COVID-19 Data Sources and Methodologies page', () => {
```
MaxwellKendall, line 17: instead of doing these nested describes, we can just do a single it block. When you execute the test it will print out the file it's testing to the console so we don't have to do this IMO.

MaxwellKendall (approved):

tests/components/covid19/DataSourcesAndMethodologiesPage-test.jsx
ebdabbs: That's a good point and I can simplify the descriptions going forward. These tests can be deleted after 3/3 so I'll leave them as-is here!

PR #2411, opened 2021-02-24
[DEV-6930] Fix Unlinked Awards column sort

ebdabbs:
**High level description:**

On the single agency submission statistics page, fixes sorting on the Unlinked Awards columns. 

**Technical details:**

Also clears the error message when a network error is followed by a successful subsequent request.

**JIRA Ticket:**
[DEV-6930](https://federal-spending-transparency.atlassian.net/browse/DEV-6930)

Author:

Reviewer(s):
- [x] Code review complete

### DIFF
--- a/src/js/containers/aboutTheData/AgencyDetailsContainer.jsx
+++ b/src/js/containers/aboutTheData/AgencyDetailsContainer.jsx
@@ -144,6 +144,7 @@ const AgencyDetailsContainer = ({ modalClick, agencyName, agencyCode }) => {
                 setRows(parseRows(res.data.results));
                 setTotalItems(res.data.page_metadata.total);
                 setLoading(false);
+                setError(false);
             }).catch((err) => {
                 if (!isCancel(err)) {
                     setError(true);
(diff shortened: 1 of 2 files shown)


src/js/containers/aboutTheData/AgencyDetailsContainer.jsx
```diff
@@ -144,6 +144,7 @@ const AgencyDetailsContainer = ({ modalClick, agencyName, agencyCode }) => {
                 setRows(parseRows(res.data.results));
                 setTotalItems(res.data.page_metadata.total);
                 setLoading(false);
+                setError(false);
```
jonhill13, line 147: We should be setting the error to false before the request before `setLoading(true);`.

e.g. If a user makes a request, and visually sees data in the table, then sorts, we make another request and the user visually sees an error, then another user makes a new request the error will still be shown while the request is in flight where we would want to visually notify the user with a loading screen that we are retrieving more data.

jonhill13 (changes requested):

jonhill13 (approved):

PR #2483, opened 2021-03-30
DEV-6954 about the data modal buttons

ebdabbs:
**High level description:**

Prevents modal buttons from displaying on cells with no data for the `Most Recent Update` and `Number of TASs Missing from Account Balance Data` columns.

**Technical details:**

Also fixes a discrepancy between "Number of **TAS** Missing from Account Balance Data" and "Number of **TASs** Missing from Account Balance Data" to match the mockup.

**JIRA Ticket:**
[DEV-6954](https://federal-spending-transparency.atlassian.net/browse/DEV-6954)

Author:
- [x] Added Unit Tests for helper functions, reducers, models and Container/Component Interactivity Expectations `if applicable` [React Testing Library](react-testing-library.md) - it's been difficult to isolate the button logic in a unit test without refactoring the containers. LMK what you think!

Reviewer(s):
- [x] Code review complete

### DIFF
--- a/src/js/containers/aboutTheData/AgenciesContainer.jsx
+++ b/src/js/containers/aboutTheData/AgenciesContainer.jsx
@@ -250,7 +250,9 @@ const AgenciesContainer = ({
         .map(({
             name: agencyName,
             code,
+            _mostRecentPublicationDate,
             mostRecentPublicationDate,
+            _discrepancyCount,
             discrepancyCount: GtasNotInFileA,
             obligationDifference,
             _gtasObligationTotal,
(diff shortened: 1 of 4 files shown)


src/js/containers/aboutTheData/AgenciesContainer.jsx
```diff
@@ -261,27 +261,31 @@ const AgenciesContainer = ({
...
-                openModal={openModal}
-                modalType="missingAccountBalance"
-                agencyData={{
-                    agencyName,
-                    gtasObligationTotal: _gtasObligationTotal,
-                    agencyCode: code,
-                    fiscalYear: selectedFy,
-                    fiscalPeriod: selectedPeriod?.id
-                }} />),
+            (mostRecentPublicationDate === '--' ?
```
brett-varney, line 264: My only concern is that if the "null" API response is ever changed, this will silently fall back to always showing the expansion icon.

brett-varney (approved):
Approved, but with a minor concern that may merit update, devs option

src/js/containers/aboutTheData/AgenciesContainer.jsx
ebdabbs: Good point, let me see if I can hook into the private/raw property from our data model instead

PR #2515, opened 2021-04-21
[tests-for-agencyv2] tests to ensure no duplicate API Requests are Made in Account Spending Section

MaxwellKendall:
**High level description:**

Changes useEffect condtions and adds tests. This is tech debt that doesnt have a ticket 🤷‍♂️

**Technical details:**

+ Unit tests to confirm only one request is made when fy is null

**JIRA Ticket:**
`N/A`

Author:
`N/A` Linked to this PR in JIRA ticket 
`N/A`Verified cross-browser compatibility: Chrome, Safari, Firefox, Internet Explorer, Edge
`N/A`Verified mobile/tablet/desktop/monitor responsiveness
`N/A`Verified that this PR does not create any *new* accessibility issues (via Axe Chrome extension)

Reviewer(s):
- [ ] Code review complete


src/js/containers/agencyV2/accountSpending/TableContainer.jsx
```diff
@@ -112,14 +112,13 @@ const TableContainer = (props) => {
         if (currentPage !== 1) {
             changeCurrentPage(1);
         }
-        else {
-            fetchSpendingByCategoryCallback();
-        }
-    }, [props.type, props.fy, props.agencyId, pageSize, sort, order, totalObligation]);
+    }, [props.type, props.agencyId, pageSize, sort, order, totalObligation]);
```
ebdabbs, line 115: When I switch tabs, change the page size, etc. _while on page 1_ the table no longer makes an API request with the new params.

ebdabbs (changes requested):

src/js/containers/agencyV2/accountSpending/TableContainer.jsx
```diff
@@ -112,14 +114,24 @@ const TableContainer = (props) => {
...
             changeCurrentPage(1);
         }
-        else {
-            fetchSpendingByCategoryCallback();
+        else if (currentPage === 1) {
+            const hasParamChanged = (
+                prevSort !== sort ||
+                prevOrder !== order ||
+                prevPage !== currentPage ||
+                prevPageSize !== pageSize
```
ebdabbs, line 122: I think we also need to check if `props.type` has changed - it's still not making a new request when I switch tabs (e.g. from Budget Functions to Program Activities)

src/js/helpers/index.js
```diff
@@ -0,0 +1,17 @@
+import { useState, useRef, useEffect } from 'react';
+
+export const useStateWithPrevious = (initialValue) => {
```
ebdabbs, line 3: This _feels_ wrong to me, I think because it is basically going back to `componentDidUpdate` but with hooks. Can you describe the use case for this? (In other words, why do we need to know the previous value of these state vars that are in the dependency array of a useEffect? It seems like all we need to know is that the value has changed, and that is all we're using this to check for in `TableContainer`.)

ebdabbs (changes requested):

src/js/helpers/index.js
MaxwellKendall: Sure. In [the react docs](https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state) they recommend this implementation. It does feel wrong but it's the official way to do it. 

Also:

> It’s possible that in the future React will provide a usePrevious Hook out of the box since it’s a relatively common use case.

The only reason we need this is because if we only check for if the value changed via the dependency array, it will fire off a duplicate request on initial render.

ebdabbs: Thanks!

ebdabbs (approved):

PR #2543, opened 2021-05-03
Meta Tag Keys

jonhill13:
> • adds keys to a few more meta tags.

- [x] Code review complete


afnazu (approved):
All good, no more key prop console error! 👌

PR #2585, opened 2021-05-26
Security Vulnerability: Bump Babel-Jest to latest and Migrate to Dev-Dependencies

jonhill13:
**High level description:**

Security Vulnerability: Bump Babel-Jest to latest and Migrate to Dev-Dependencies

- [x] Code review complete


tests/containers/agencyV2/AgencyContainerV2-test.jsx
```diff
@@ -76,3 +57,22 @@ test('an API request is made for the agency code in the URL', () => {
     });
 });
 
+// TODO - Commenting this out since it is failing our tests in the travis run even though tests are green, the rejection is causing an issue
```
ebdabbs, line 60: Are we expecting this to work with Travis at some point in the future? If not, I think we should delete this instead of commenting it out.

jonhill13: Given this is the only test we have checking for a failing API call, we certainly could remove it. I see a lot of benefit though in being able to mock failing api calls and why i just commented it out.

ebdabbs: Leaving as an example and reminder to look into it!

ebdabbs (approved):

PR #2715, opened 2021-09-03
DEV-7836 ARP filter - recipient section

ebdabbs:
**High level description:**

Filters the Covid-19 Profile Award Spending by Recipient section data based on the `publicLaw` URL query param.

**Technical details:**

- Store a new list of Covid 19 Profile DEF codes in Redux (since existing `defCodes` are used elsewhere on the site too)
  - this is a flat list of just the string for the code, e.g. `(["L", "M", "N", "O", "P", "U", "V"])`
- Update this list when the `publicLaw` query param changes (in Covid19Container.jsx)
  - `all`: every DEFC returned by useDefCodes hook
  - `american-rescue-plan`: use the defcByPublicLaw mapping
- Update the places we use existing `defCodes` to create API params to use this new list instead

Also changed file names for clarity
- SpendingByRecipient --> SpendingByRecipientContainer (moved this to `/container` directory from `/component`)
- SpendingByRecipientContainer --> RecipientTableContainer
- RecipientContainer --> RecipientSection (moved this to `/component` directory from `/container`)

**JIRA Ticket:**
[DEV-7836](https://federal-spending-transparency.atlassian.net/browse/DEV-7836)

Author:

Reviewer(s):
- [x] Code review complete


src/js/containers/covid19/recipient/RecipientTableContainer.jsx
```diff
@@ -0,0 +1,347 @@
...
+                    rowData.awardCount
+                ]]);
+            setParsedRows(rowsWithUnlinked);
+        }
+        else {
+            setParsedRows(rows);
+        }
+    };
+
+    const fetchSpendingByRecipientCallback = useCallback(() => {
```
brett-varney, line 219: @ebdabbs I see this warning in ESlint:

React Hook useCallback does nothing when called with only one argument. Did you forget to pass an array of dependencies?

brett-varney (changes requested):
question, not necessarily change request

brett-varney:
After a couple of test runs, it started calling APIs with no def codes (empty array). I'll poke around to see if I can see why.

brett-varney (changes requested):
Seems to work, but I'm concerned about the call to /recipients/count with no defcodes. It isn't happening before this update.

src/js/containers/covid19/recipient/RecipientTableContainer.jsx
brett-varney: But now I see that this was there before.

ebdabbs: Removing `useCallback` since I don't think it's necessary here.

ebdabbs:
Added a check for the empty DEFC array to `SpendingByRecipientContainer`

brett-varney (approved):

PR #2853, opened 2022-01-04
[DEV-8052] Status of funds click through functionality

afnazu:
**High level description:**

As a user, I want to be able to drill down into a sub-component to see data broken down by federal account.

**Technical details:**
Updating parent state from child to make d3 fetch and rerender on click

**JIRA Ticket:**
[DEV-8052](https://federal-spending-transparency.atlassian.net/browse/DEV-8052)

**Mockup:**
https://bahdigital.invisionapp.com/share/BSIAHSLA9PE#/screens

Author:
- [ ] Linked to this PR in JIRA ticket
- [ ] Verified cross-browser compatibility: Chrome, Safari, Firefox, Internet Explorer, Edge
- [ ] Verified mobile/tablet/desktop/monitor responsiveness
- [ ] Verified that this PR does not create any *new* accessibility issues (via Axe Chrome extension)
- [ ] Added Unit Tests for helper functions, reducers, models and Container/Component Interactivity Expectations `if applicable` [React Testing Library](react-testing-library.md)
- [ ] [API contract](https://github.com/fedspendingtransparency/usaspending-api/tree/dev/usaspending_api/api_contracts) updated `if applicable`
- [ ] [Component Library Integration Status Report](https://github.com/fedspendingtransparency/data-act-documentation/blob/data-transparency-ui/frontend_apps/component-library-integration-status.md) updated `if applicable`

Reviewer(s):
- [ ] Code review complete

### DIFF
--- a/src/js/components/agencyV2/statusOfFunds/StatusOfFunds.jsx
+++ b/src/js/components/agencyV2/statusOfFunds/StatusOfFunds.jsx
@@ -35,6 +35,10 @@ const StatusOfFunds = ({ fy }) => {
     const [results, setResults] = useState([]);
     const { overview, selectedSubcomponent } = useSelector((state) => state.agencyV2);
 
+    const updateResults = (resData) => {
+        setResults(resData);
+    };
+
     useEffect(() => {
         if (request.current) {
             request.current.cancel();
@@ -114,7 +118,7 @@ const StatusOfFunds = ({ fy }) => {
                         selectedSubcomponent={selectedSubcomponent} />
                 </FlexGridCol>
                 <FlexGridCol className="status-of-funds__visualization" desktop={9}>
-                    { results.length !== 0 ? <VisualizationSection level={level} agencyId={overview.toptierCode} agencyName={overview.name} fy={fy} results={results} /> : <LoadingMessage /> }
+                    { results.length !== 0 ? <VisualizationSection level={level} agencyId={overview.toptierCode} agencyName={overview.name} fy={fy} results={results} updateResults={updateResults} /> : <LoadingMessage /> }
                     <Pagination
                         currentPage={currentPage}
                         changePage={changeCurrentPage}
--- a/src/_scss/pages/agencyV2/statusOfFunds/_visualizationSection.scss
+++ b/src/_scss/pages/agencyV2/statusOfFunds/_visualizationSection.scss
@@ -12,6 +12,11 @@
             height: 100% !important;
             width: 100% !important;
         }
+        .parent-g { visibility: hidden; rect#hlines { fill-opacity: 0; }}
+        .parent-g > * { visibility: visible; rect#hlines { fill-opacity: 0; }}
+        .parent-g > * { transition: opacity 150ms ease-in-out 100ms; }
+        .parent-g:hover > * { opacity: 0.4; }
+        .parent-g > *:hover { cursor: pointer; opacity: 1; transition-delay: 0ms, 0ms; rect#hlines { filter: drop-shadow(0px 3px 3px rgba(0, 0, 0, 0.4)); fill-opacity: 1; stroke: none }}
         text.y-axis-labels {
             font-size: rem(20);
             @media(max-width: 1499px) {
--- a/src/js/apis/agencyV2.js
+++ b/src/js/apis/agencyV2.js
@@ -52,3 +52,7 @@ export const fetchAgencySlugs = () => apiRequest({
 export const fetchSubcomponentsList = (code, fy, page) => apiRequest({
     url: `v2/agency/${code}/sub_components/${fy ? `?fiscal_year=${fy}` : ''}${page ? `&page=${page}` : ''}`
 });
+
+export const fetchFederalAccountsList = (code, slug, fy, page) => apiRequest({
+    url: `v2/agency/${code}/sub_components/${slug}/${fy ? `?fiscal_year=${fy}` : ''}${page ? `&page=${page}` : ''}`
+});
--- a/src/js/components/agencyV2/statusOfFunds/VisualizationSection.jsx
+++ b/src/js/components/agencyV2/statusOfFunds/VisualizationSection.jsx
@@ -14,19 +14,21 @@ const propTypes = {
     agencyId: PropTypes.string,
     agencyName: PropTypes.string,
     fy: PropTypes.string,
-    results: PropTypes.array
+    results: PropTypes.array,
+    updateResults: PropTypes.func
 };
 
 const VisualizationSection = ({
     level,
     agencyName,
     fy,
-    results
+    results,
+    updateResults
 }) => (
     <div className="status-of-funds__visualization">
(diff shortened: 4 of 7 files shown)


anjenkin (changes requested):
Looks good!  A couple things, after going to page 2 (or greater) when you select a sub-component, there is a bit of a delay before the federal accounts show. Also, the keyboard accessibility ACs are not passing - 

* Each group of bars (obligated and TBR) is keyboard accessible

* When clicking, hitting “Enter”, or tapping the group of bars, the chart is updated to show federal account data for the selected Sub-Component

src/js/components/agencyV2/statusOfFunds/StatusOfFunds.jsx
```diff
@@ -87,19 +133,20 @@ const StatusOfFunds = ({ fy }) => {
...
 
-    // TODO - remove mock data when DEV-8052 is implemented
-    const mockData = {
-        name: "Bureau of the Census",
-        id: "bureau_of_the_census",
-        total_budgetary_resources: 5000000,
-        total_obligations: 3000000.72
+    const onClick = (selectedLevel, data) => {
+        // reset to page 1 on drilldown
+        changeCurrentPage(1);
```
anjenkin, line 138: This seems to be causing the issue with the delayed update mentioned.   New data is fetched and the chart is updated here in response to the page change, but also get's updated due to the subcomponent change that happens later in this function.

src/js/components/agencyV2/statusOfFunds/StatusOfFunds.jsx
```diff
@@ -72,13 +75,56 @@ const StatusOfFunds = ({ fy }) => {
...
+            fetchFederalAccounts(subcomponent);
+        }
+    }, [subcomponent]);
+
     useEffect(() => {
-        const hasParamChanged = (
-            prevPage !== currentPage || prevPageSize !== pageSize
-        );
-        if (hasParamChanged) {
+        if (prevPage !== currentPage && level === 0) {
```
anjenkin, line 122: If you add the following condition to prevent fetching new data in this useEffect when !(prevPage > 1 && currentPage === 1) and add setLoading(true) in the onClick function that fixes the issues. I'm not sure if this works for all cases and it seems like there should be a better way to check !(prevPage > 1 && currentPage === 1) such as setting a new state to transitioning and to prevent fetching new data here when the transitioning state is set.

afnazu: Thanks Andrea I pushed a fix adding the check for currentPage !== 1.  I double checked the tab/enter functionality and it is still working on my end, the tab through starts with the x axis labels then goes through each bar group- are you not able to tab through the chart at all? @anjenkin

anjenkin: I see tabbing works.  I was expecting a different tab order. I was thinking you could tab to the label and then the next tab would set focus to the bar next to the label.

There's an issue with checking currentPage !== 1 because that breaks pagination if you select a page greater than 1 and then going back to page 1.

I just tested the following and it works if you set state while the chart is resetting in this case.  So in the useEffect that responses to the currentPage changing then if you can check for the "resetting state" and don't fetch the data if resetting.  This is a tricky visualization!


```
    const [resetBarCnt, setResetBarCnt] = useState(false);

    useEffect(() => {
        if (resetBarCnt) {
            setResetBarCnt(false);
        } else {
            if (prevPage !== currentPage && level === 0) {
                fetchAgencySubcomponents();
            }
            if (prevPage !== currentPage && level === 1) {
                fetchFederalAccounts(subcomponent);
            }
        }
    }, [currentPage]);

    useEffect(() => {
        if (resetBarCnt) {
            setLoading(true);
            changeCurrentPage(1);
        }
    }, [resetBarCnt]);

    const onClick = (selectedLevel, data) => {
        // reset to page 1 on drilldown
        setResetBarCnt(true);
        const subcomponentTotalData = Object.create(BaseStatusOfFundsLevel);
        subcomponentTotalData.populate(data);
        dispatch(setSelectedSubcomponent(subcomponentTotalData));
        setSubcomponent(subcomponentTotalData);
    };
```

afnazu: Ah thank you good catch on the behavior when changing to page 1 from another page > 1 I had missed that condition when testing - I updated the PR with your page change state logic.
Re: tabbing, I have set a tab-index of 0 to both the bar groups and y axis labels so the tab indexes are grouped by d3 element

anjenkin: Geez, I found one more issue while testing. If you click on a sub-component on page 1 (which doesn't trigger a page change), click back, and then try to paginate, the data doesn't update because we didn't trigger a page change and the resetPageChange state isn't cleared.  I think we should be good just setting resetPageChange state to false in the goBack function.  I can't think of another case where we would have an issue.

afnazu: Ahh setting page change to false in onClick resolved both issues for me - just pushed updates let me know if you are still seeing the issue

anjenkin: Yeah this seems to work - 
`    useEffect(() => {
        if (resetPageChange) {
            setLoading(true);
            if (currentPage === 1) {
                setResetPageChange(false);
            } else {
                changeCurrentPage(1);
            }
        }
    }, [resetPageChange]);

    useEffect(() => {
        if (fy && overview.toptierCode) {
            fetchAgencySubcomponents();
        }
    }, [fy, overview.toptierCode]);

    const onClick = (selectedLevel, data) => {
        // reset to page 1 on drilldown
        setResetPageChange(true);
        const subcomponentTotalData = Object.create(BaseStatusOfFundsLevel);
        subcomponentTotalData.populate(data);
        dispatch(setSelectedSubcomponent(subcomponentTotalData));
        setSubcomponent(subcomponentTotalData);
    };`

afnazu: Thanks this seems to work on my end too - I included the same logic to resolve a flickering issue with the goBack() which fixed that as well.. fingers crossed that this is the golden fix 🙏

anjenkin: This works but if I go to the Dept of Treasury V2 page, click the Departmental Offices sub-component and go to page 2, the bars are the wrong height and it's only showing 5 bars.  However, if I go to page 3 and 4, the bars show correctly
Screen Shot 2022-01-11 at 12 18 47 PM
.

afnazu: @anjenkin Thanks for that catch! The way we were truncating texts caused some y axis labels to have the same value which causes d3 to put them in the same place. Just pushed an update so that we are now doing the truncation after getting the full subcomponent name so this resolves the bar heights issue

PR #2862, opened 2022-01-10
8301 - Ensure tooltip closes when mousing out from the donut chart

anjenkin:
**High level description:**

Ensure tooltip closes when mousing out from the donut chart.

**Technical details:**

The hit area for the outer ring is so small so the mouseout event to close the tooltip wasn't always getting captured.  I added a padding around the donut so if the user hovers on or over the padding, the tooltip will close.

**JIRA Ticket:**
[DEV-8301](https://federal-spending-transparency.atlassian.net/browse/DEV-8301)

Author:

Reviewer(s):
- [x] Code review complete


brett-varney:
@anjenkin I can't duplicate the problem in QAT; maybe it was already fixed somehow? Do you have an example besides the one in the ticket (which I don't see how to duplicate)?

brett-varney (approved):

PR #2898, opened 2022-01-27
Fix/ie break react checkboxes

anjenkin:
High level description:

Currently IE is showing as a blank screen in the QAT environment (though it works locally). The issue is in the vendors bundle so we suspect the problem is related to one of the package updates. In this PR I reverted the react-checkbox-tree version increase.

- [ ] Code review complete


package.json
```diff
@@ -5,13 +5,13 @@
     "repository": "fedspendingtransparency/usaspending-website",
     "main": "src/index.js",
     "scripts": {
-        "start": "webpack-dev-server --open --history-api-fallback --progress --config ./webpack/webpack.dev.config.js",
+        "start": "webpack-dev-server --open --history-api-fallback --progress --config ./webpack/webpack.prod.config.js --json > stats.json",
```
brett-varney: Is this "prod" left over from your local testing?

stats.json
```diff
@@ -0,0 +1,38 @@
+Commit Hash for this build:  5d254f037531b041b7e6ee7894812b4bfd405b96
```
brett-varney: Should this file be included, or is it temp for testing?

brett-varney (approved):

PR #3084, opened 2022-05-23
8616 - added names without links to top 5 when no slug is available

anjenkin:


nick-torres (approved):
changes are as expected

PR #3093, opened 2022-05-24
Remove mapbox key

anjenkin:


sethstoudenmier (approved):
LGTM, just changing back to what it was previously.

PR #3410, opened 2023-01-03
Hotfix/9402 - Word of the Day showing as blank fix Env Promotion

anjenkin:


nick-torres:
shouldn't this be from staging?

nick-torres (approved):

PR #3982, opened 2024-02-15
10520 - Advanced search map selection and zoom to State and country

anjenkin:
**High level description:**

When a single country, state, county, and CD location filter is set, the map should update to show that location selected.  For now we're using the state center point for county and congressional district instead of getting the county and congressional district centers.  We'll get these center points in a separate ticket. (See the technical details below for further details).

Also, with this ticket fixed an issue that broke getting the data in view on zoom in / out.  this was a regression from one of the map component class -> functional component tickets.

**Technical details:**

This ticket was large for a number of reasons.
- had to convert src/js/components/search/visualizations/geo/GeoVisualizationSection.jsx to functional component
- had to integrate updates from mapwrapper and geovisualizationsectioncontainer class -> functional component tickets
- there were several issues with components losing state which need to be addressed in a future ticket; it's happening so frequently in these components now and i don't think using refs the way we're using them is the correct solution (ie. do we need to use redux for map state including data, maplayer, scope, etc).
- there are three different ways needed to determine the center point - country needs to be done by Mapbox geolocation api, state can use the state center maps that were already in our code, county and congressional district need to get the center lat and long from the Mapbox tile.   however, the mapbox tileset is selected after we've set the center point.  As a result, the fix for now is to use the state center point for county and congressional district instead of getting the county and congressional district centers.  We'll get these center points in a separate ticket.
- probably need to make sure nothing was done to break the state profile map 

**JIRA Ticket:**
[DEV-10520](https://federal-spending-transparency.atlassian.net/browse/DEV-10520)

Author:

Reviewer(s):
- [ ] Code review complete


nick-torres:
image

src/js/containers/search/visualizations/geo/GeoVisualizationSectionContainer.jsx
```diff
@@ -254,28 +260,52 @@ const GeoVisualizationSectionContainer = (props) => {
...
+        }
+    };
+
     const mapScopeLogic = (type) => {
         const selectedLocationByType = type === "pop" ? "selectedLocations" : "selectedRecipientLocations";
         // there is only 1 item, place of performance
         if (props.reduxFilters[selectedLocationByType].size === 1) {
             const onlyObject = props.reduxFilters[selectedLocationByType].first().filter;
+            setSingleLocationSelected(onlyObject);
             if (onlyObject.district_current || onlyObject.district_original) {
```
nick-torres, line 294: should these all be onlyObject with a null check

ex onlyObject?.district_current

anjenkin: I don't think so.  The `if (onlyObject.district_current || onlyObject.district_original)` condition was added for a previous ticket and not touched for this ticket.  So this condition is already in prod.

nick-torres (approved):

PR #4065, opened 2024-04-24
downmerge to QAT

anjenkin:


src/_scss/layouts/default/default.scss
```diff
@@ -1,3 +1,4 @@
 // Default Layout: The default layout only includes the header and footer styles
 @import "./header/header";
+@import "./header/warning";
```
nick-torres, line 3: was there a hotfix in staging that we forgot to downmerge to qat??

nick-torres (approved):

PR #4079, opened 2024-05-09
Resolving merge conflict between QAT and Staging

anjenkin:


src/js/components/sharedComponents/header/Header.jsx
```diff
@@ -1,21 +1,20 @@
...
-        super(props);
-
-        // bind functions
-        this.skippedNav = this.skippedNav.bind(this);
-    }
+const Header = () => {
+    const location = useLocation();
 
-    skippedNav(e) {
+    console.log(location.pathname);
```
nick-torres, line 16: should this be deleted?

PR #4223, opened 2024-08-15
11292 - Adjusted the homepage jump to location for the map

anjenkin:
https://data-transparency-bfs.slack.com/archives/CE5P5AWE8/p1723724473481859?thread_ts=1723657841.782449&cid=CE5P5AWE8

### DIFF
--- a/src/js/components/search/newResultsView/SearchSectionWrapper.jsx
+++ b/src/js/components/search/newResultsView/SearchSectionWrapper.jsx
@@ -113,7 +113,7 @@ const SearchSectionWrapper = ({
         else if (matchedSection === 'time') {
             rectTopOffset = 1680;
         }
-        else if (matchedSection === 'awards') {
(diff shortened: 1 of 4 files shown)


brianpetway:
When clicking the links from the homepage, the scrollTo does go to that section on the results page initially, but then it jumps to a different slightly different place in the page after the component loads. This happens for map, time, and categories sections.
Video attached.


https://github.com/user-attachments/assets/28f8c008-9f30-4c33-a576-c34d63190198

brianpetway:
When scrolling down the results page - without going to the page from a homepage link - the page jumps to a different place after the new section has loaded, you see this for the time and categories sections.
And the map section now loads before it is scrolled into view, I think because it has no Placeholder component.

https://github.com/user-attachments/assets/7f4fbed8-fd5b-4093-b7ad-6f3f542c69ba

PR #5230, opened 2026-04-15
new awards is removed when there are no time periods selected

jagu65:
see title


anjenkin (changes requested):
Encountered a bug - 
1. add 2025 
2. check new award only
3. submit
4. remove the new award only and submit
5.then attempt to remove 2025, 
result:  the screen goes blank.

This happens when both submitting from the side panel or the active filter ui

anjenkin (approved):

PR #5354, opened 2026-06-11
fix homepage left card

nick-torres:
**Description:**

transformed titled did not match the slug so the card image wasn't showing and the link was going to a 404


jagu65 (dismissed):

src/js/helpers/homepageFeaturedContentHelper.js
```diff
@@ -11,9 +11,9 @@ const partition = (array, isValid) => array.reduce(
 const fallbackMarketingArticle = {
     title: "Exploring America's Finances Has Never Been Easier",
```
anjenkin: the title is wrong "Has Never Been Easier" needs to be removed here too

anjenkin (changes requested):

anjenkin (dismissed):

jagu65 (approved):

